PR #4, opened 2018-06-08
feat: Use status code 204 in place of 404.

jasonmit:


packages/@pollyjs/node-server/src/api.js
```diff
@@ -1,5 +1,5 @@
-import path from 'path';
 import fs from 'fs-extra';
+import path from 'path';
```
jasonmit: My editor is doing something weird, need to fix it.  Keeps reorganizing imports on save.

PR #32, opened 2018-06-21
fix: Rest server on Windows

jasonmit:


packages/@pollyjs/node-server/src/express/register-api.js
```diff
@@ -1,13 +1,20 @@
...
+    return slash;
+  }
+
+  return `/${slash}`;
+}
+
 export default function registerAPI(app, config) {
   config = { ...DefaultConfig, ...config };
-  config.apiNamespace = path.join('/', config.apiNamespace);
+  config.apiNamespace = prependSlash(config.apiNamespace);
```
jasonmit: Ends up being `"\" + config.apiNamespace` on Windows because of `path.join`.

PR #36, opened 2018-06-21
fix: Ensure polly's middleware goes before ember-cli's

cibernox:
This should allow polly to work nicely with ember-cli's when the
`--proxy` option is passed.

### DIFF
--- a/packages/@pollyjs/ember/package.json
+++ b/packages/@pollyjs/ember/package.json
@@ -76,6 +76,9 @@
     "node": "6.* || >= 7.*"
   },
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "before": [
+      "proxy-server-middleware"
+    ]
   }
 }


Turbo87:
FYI this doesn't work quite as expected. The `--proxy` option in Ember CLI is only used for `ember serve`, so it does work when visiting `/tests`, but not when using `ember test`. That means that requests to a relative URL like `/api/foo` will **not** be proxied in test mode. What I did instead is this:
```js
server.any().on('request', req => {
  if (req.url.startsWith('/api')) {
    req.url = `https://skylines.aero${req.url}`;
  }
});
```

This solution seems to work quite well now. I'm assuming that this is/will be a common issue for people using Polly.js, so it might make sense to add it to the documentation somewhere.

cibernox:
Just to clarify, this dies works as intended as long as Polly recorded the events already.

Turbo87:
possibly, I wasn't aware testem could do this 🙈

offirgolan:
😄 lemme know if it works and I'll add some documentation around it!

Turbo87:
indeed, this seems to work very well. thanks for the idea!

cibernox:
I've also used that config and it's great. I didn't know that trick either.

Issue #28, opened 2018-06-18
feature: IE11 support


piotrpalek:
Hey, I've been trying out pollyjs in an Ember.js app, and noticed that the app won't work at all because the imported pollyjs files include statements like arrow functions which don't work with IE.

The issue seems to be that the Polly.js addon uses `app.import` to include the libraries, but afaik `app.import` doesn't run the code through webpack, so it doesn't transform it into viable ES5 code. It just includes the code verbatim.

edit: btw sidenote, shouldn't pollyJS be included only when running in the test environment by default?

jasonmit:
Currently not a browser we target.

> edit: btw sidenote, shouldn't pollyJS be included only when running in the test environment by default?

Dev and testing.  Prod it should not be included.

jasonmit:
@piotrpalek we're going to update Polly to support down to IE11.  Likely will happen in the next release.

piotrpalek:
@jasonmit cool, I've currently sidestepped the problem by only including it when needed (testing basically) but :+1:

Ekman:
@jasonmit any status updates on this?

jasonmit:
@Ekman no update, I haven't revisited after fixing Edge.  We also need to setup a Windows CI environment.

offirgolan:
Closing this issue. Feel free to reopen if IE11 support is still needed.

PR #37, opened 2018-06-22
chore: Tests now run in MS Edge

jasonmit:
This PR does not include the changes to rollup since I consider that separate for Edge support.  IE 11 is a bit more work, but I have most of that done in another branch just need to verify it.

I'll probably introduce a test:legacy command since it's going to slow the build down a bit and make normal debugging in modern browsers difficult.

One of two changes required for #28


packages/@pollyjs/core/src/server/route.js
```diff
@@ -3,13 +3,20 @@ import Handler from './handler';
...
+        return true;
+      },
       get(source, prop) {
         if (prop === 'params') {
           // Set the request's params to given route's matched params
           return route.params;
         }
 
-        return source[prop];
+        return Reflect.get(source, prop);
```
offirgolan: Why the change to Reflect.get here?

PR #38, opened 2018-06-22
fix: If the source URL is relative the normalized must not have domain

cibernox:
Description of the symptoms of bug and the fix:

1. I run and recorded requests when I my tests in `http://localhost:3000`
2. I then pushed the branch to CI and it failed. It turns out that the recorded requests were not being replayed when ember-cli runs in CI.
3. After some debugging I discovered that the id of the request (the md5 generated from headers, url and method) is different because the normalized URL is different (`/my-url` was being normalized to `http://localhost:7357/my-url`)

The fix is to, on the URL normalization step, if the source url doesn't have a domain, the normalized version should not contain a domain either, effectively keeping the URL relative.

**Warning**: This is a breaking change. Recordings of relative urls that have already been captured will have to be re-recorded.

### DIFF
--- a/packages/@pollyjs/core/src/utils/normalize-request.js
+++ b/packages/@pollyjs/core/src/utils/normalize-request.js
@@ -1,21 +1,23 @@
 import URL from 'url-parse';
 import removeHostFromUrl from './remove-host-from-url';
 import isObjectLike from 'lodash-es/isObjectLike';
+import isAbsoluteUrl from 'is-absolute-url';
 
 const { keys } = Object;
-
-function isAbsoluteUrl(url) {
-  /^[a-z][a-z0-9+.-]*:/.test(url);
-}
 
 export function method(method) {
   return (method || 'GET').toUpperCase();
 }
 
 export function url(url, config = {}) {
-  let isRelative = !isAbsoluteUrl(url);
   const parsedUrl = new URL(url, true);
 
+  // Remove the host if the url is relative
+  if (!isAbsoluteUrl(url)) {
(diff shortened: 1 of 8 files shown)


packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -8,6 +8,7 @@ export function method(method) {
 }
 
 export function url(url, config = {}) {
+  let isRelative = url[0] === '/';
```
offirgolan: Lets use this: https://github.com/sindresorhus/is-absolute-url

packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -27,7 +28,11 @@ export function url(url, config = {}) {
     parsedUrl.set('query', sortedQuery);
   }
 
-  return parsedUrl.href;
+  if (isRelative) {
```
offirgolan: Lets put this behind a config option. We can enable it by default but at least users will have the ability to turn this functionality off. 

```js
{ 
  matchRequestsBy: { 
    url: { 
      relative: true 
    } 
  } 
}
```

_config name not final_

packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -27,7 +28,11 @@ export function url(url, config = {}) {
...
+    return parsedUrl.href.replace(parsedUrl.origin, '');
```
offirgolan: We have a util for this `utils/remove-host-from-url`

PR #40, opened 2018-06-22
feat: Make recording size limit configurable

jasonmit:


packages/@pollyjs/cli/bin/cli.js
```diff
@@ -25,6 +25,11 @@ cli
     'recordings directory',
     Polly.Defaults.recordingsDir
   )
+  .option(
```
jasonmit: Unverified

PR #44, opened 2018-06-26
fix(core): Set `url` on the fetch Response object

### LINKED ISSUE #43: Cannot set/use Response URL using client-side server
I have a piece of code that makes two requests in parallel. When I have both responses I use the URL of the response objects to decide how to handle the response data.

I've set up Polly using a client-side server but the response URLs are coming back as empty strings. I have tried a couple of different ways of setting the URL in my generated response objects but nothing seems to work. I am currently digging into the source code to see if I can identify a method of achieving the results I need but I so far haven't found anything.

Am I missing something? I can try and provide a reduced test case if needed. 
Is this something that you would consider adding or accepting a PR for?

Thanks!

offirgolan:
Resolved #43.

### DIFF
--- a/packages/@pollyjs/core/src/adapters/fetch/index.js
+++ b/packages/@pollyjs/core/src/adapters/fetch/index.js
@@ -84,7 +84,7 @@ export default class FetchAdapter extends Adapter {
     defineProperty(fetchResponse, 'url', {
       /*
         Since `url` can be a relative url and Response always has an absolute
-        one, use URL to stitch together the origin if necessary.
+        one, use URL to attach the host if necessary.
       */
       value: new URL(url).href
     });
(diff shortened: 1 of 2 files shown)


packages/@pollyjs/core/src/adapters/fetch/index.js
```diff
@@ -68,12 +70,26 @@ export default class FetchAdapter extends Adapter {
...
     });
+
+    /*
+      Response does not allow `url` to be set manually (either via the
+      constructor or assignment) so force the url property via `defineProperty`.
+    */
+    defineProperty(fetchResponse, 'url', {
+      /*
+        Since `url` can be a relative url and Response always has an absolute
+        one, use URL to stitch together the origin if necessary.
```
jasonmit: Unsure what stitch together the origin means

offirgolan: Oh my bad, I meant that it attaches the host name to a relative url. I'll update the comment.

Issue #19, opened 2018-06-14
Jest Support?


yarivkenan:
Hi looks like a great project, can I use it with Jest?

enesTufekci:
@yarivkenan Why not? Have you tried it yet?

yarivkenan:
I'll try and get back to you.

RemeJuan:
Jest is for unit testing, not Journey testing, which is where this would come in handy.

webstercharly:
@RemeJuan Jest is a test runner, it is not solely limited to Unit testing, it can be used for any level of the pyramid.

I actually came here looking to see compatibility for a project which is using Jest with Puppeteer.

klarkc:
Actuallly I am refactoring a jest suite for everything (e2e, unit, integration), I will do some try with pollyjs in replace of supertest and nock. If I can replace both, it will be a double win

harunurhan:
@webstercharly that is exactly what I am trying to do and I bumped into #41

webstercharly:
@harunurhan yep exactly the same as me. Couldn’t solve it so I ended up intercepting the requests, using regex on url and then returning request with my json.

Shame, would have loved to use polly and not implement my own

harunurhan:
@webstercharly Ok, I guess I'll have to keep my custom solution too.

offirgolan:
As of right now, Polly is a client side library (meaning it will only run in the browser) but we are working towards an isomorphic solution!

I'll be working on this in the next few weeks to hopefully find an interim solution to Jest. I'll update this thread when I make some headway.

harunurhan:
Probably It doesn't work for jest running on puppeteer environment.

Have you tried it already @webstercharly ?

offirgolan:
@harunurhan can you clarify as to what you mean by "jest running on puppeteer environment"? Are you talking about just running Puppeteer and testing with Jest (as documented [here](https://jestjs.io/docs/en/puppeteer)) or with something like [jest-environment-puppeteer](https://www.npmjs.com/package/jest-environment-puppeteer)?

webstercharly:
Using Jest as a runner and Jests assertion library, with the specs initialising and executing puppeteer commands.

I have a setup using Jest environment package.

I didn’t give it a try yet, don’t need it anymore but would like to @harunurhan

harunurhan:
@offirgolan Sorry for the confusion, my setup is similar to the first one (`jest-puppeteer`).

offirgolan:
[...]

PR #50, opened 2018-06-29
CORs passthrough on node-server

jasonmit:
Related to comments in #19 

Address the nuance of having to pass in the testURL as the hostname of the node server.  Now testURL can be anything (other than the default `about:blank`), we recommend `http://localhost` to be the default.

### DIFF
--- a/packages/@pollyjs/node-server/src/server.js
+++ b/packages/@pollyjs/node-server/src/server.js
@@ -36,7 +36,9 @@ export default class Server {
     this.server = this.app
       .listen(port, host)
       .on('listening', () => {
-        console.log(`Listening on http://${host || 'localhost'}:${port}/\n`);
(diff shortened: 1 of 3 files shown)


packages/@pollyjs/node-server/src/server.js
```diff
@@ -10,6 +11,7 @@ export default class Server {
   constructor(config = {}) {
     this.config = { ...DefaultConfig, ...config };
     this.app = express();
+    this.app.use(cors());
```
jasonmit:
```
{
  "origin": "*",
  "methods": "GET,HEAD,PUT,PATCH,POST,DELETE",
  "preflightContinue": false,
  "optionsSuccessStatus": 204
}
```

is the default config.  Covers what we need here.

PR #52, opened 2018-07-01
feat: Class events and EventEmitter

offirgolan:
- Adds `Polly.on('create')` and `Polly.on('stop')` events
- Refactor eventing to an EventEmitter class

### DIFF
--- a/packages/@pollyjs/core/src/-private/event-emitter.js
+++ b/packages/@pollyjs/core/src/-private/event-emitter.js
@@ -26,7 +26,7 @@ function assertListener(listener) {
   );
 }
 
-export default class Evented {
+export default class EventEmitter {
   constructor({ eventNames = [] }) {
     this[EVENTS] = new Map();
     this[EVENT_NAMES] = new Set(eventNames);
@@ -93,14 +93,14 @@ export default class Evented {
   }
 
   async emit(eventName, ...args) {
-    await Promise.all(
-      this.listeners(eventName).map(listener => listener(...args))
-    );
-  }
-
-  async emitSerial(eventName, ...args) {
     for (const listener of this.listeners(eventName)) {
       await listener(...args);
     }
+  }
+
+  async emitParallel(eventName, ...args) {
+    await Promise.all(
+      this.listeners(eventName).map(listener => listener(...args))
+    );
   }
 }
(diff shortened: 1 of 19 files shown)


packages/@pollyjs/core/src/server/handler.js
```diff
@@ -1,70 +1,34 @@
...
-    );
-
-    const events = this.get(EVENTS);
-
-    if (!events.has(eventName)) {
-      events.set(eventName, []);
-    }
-
-    events.get(eventName).push(handler);
+  on() {
```
jasonmit: pass the args in, helps with readability and understanding what is going on

packages/@pollyjs/core/src/-private/event-emitter.js
```diff
@@ -0,0 +1,218 @@
...
+}
+
+function assertListener(listener) {
+  assert(
+    `Invalid listener provided. Expected function, received: "${typeof listener}".`,
+    typeof listener === 'function'
+  );
+}
+
+export default class EventEmitter {
```
jasonmit: Does anything event emitters out there implement a similar API?  https://github.com/sindresorhus/emittery is one that I know of that may

offirgolan: Yeah I found that one too but it was lacking the ability to:

1. Get all the listeners for an event
2. Emit events synchronously 
3. Pass more than one argument when emitting an event (which would break the api)

packages/@pollyjs/ember/addon/-private/preconfigure.js
```diff
@@ -1,15 +1,17 @@
 import { Polly } from '@pollyjs/core';
 
-Polly.configure({
```
jasonmit: Can removed static configure() again

packages/@pollyjs/core/src/-private/event-emitter.js
jasonmit: Fair points, lets continue with what you have.  One nit, I think we should rally around async vs supporting both.  Makes the API a bit confusing since depending on the caller, your handler may fire async or sync.

I see the one place that it is sync is with the create event.  This makes sense since you're not really able to await there.  But we can throw if the consumer is returning a promise there which can be detected at the end of the constructor if the promise is still not settled (Promise.race to detect that).

offirgolan: I can actually detect that from the emitSync method. I can check if the return value is promise like and throw if it is.

jasonmit: But emitSync sticks around and over time could get confusing.  "Is this an async event or sync event?"

PR #54, opened 2018-07-02
fix: Changes self to global, rollup-plugin-node-globals makes isomorphic

jasonmit:


packages/@pollyjs/core/src/persisters/rest/ajax.js
```diff
@@ -1,6 +1,6 @@
 const { keys } = Object;
-const REQUEST_ASYNC = !/PhantomJS/.test(self.navigator.userAgent);
-const NativeXMLHttpRequest = self.XMLHttpRequest;
+const REQUEST_ASYNC = !('navigator' in global) || !/PhantomJS/.test(global.navigator.userAgent);
```
jasonmit: The assumption here was if dealing with node, whatever polyfill being used wouldn't have this nuance that was being guarded here

packages/@pollyjs/core/src/persisters/rest/ajax.js
```diff
@@ -1,6 +1,7 @@
 const { keys } = Object;
-const REQUEST_ASYNC = !/PhantomJS/.test(self.navigator.userAgent);
-const NativeXMLHttpRequest = self.XMLHttpRequest;
+const REQUEST_ASYNC =
```
jasonmit: Assumes whatever polyfill being used for XHR in node would support the nuance being worked around within phantom.

PR #57, opened 2018-07-03
feat: Abort and passthrough from an intercept

### LINKED ISSUE #48: Conditional request interception and shared request recordings
In our app we have many tests that make similar, or in some cases identical requests. Examples of such request are user and config data used by the app. 

We can manually pull down this data and save it as a fixture that we load and use during a request interception but it would be great if there was a way to designate some requests as shared between recordings (using some kind tag or key). 

Another option would be to have the capability to intercept a request and have some API for breaking out of the interception (let Polly handle it as usual). 

I guess these are two rather different ideas but I think they might both be useful.

offirgolan:
Resolves #48.

```js
server
  .get('/ping')
  .intercept((req, res, interceptor) => interceptor.abort());

server
  .get('/pong')
  .intercept((req, res, interceptor) => interceptor.passthrough());
```

### DIFF
--- a/packages/@pollyjs/adapter/src/-private/interceptor.js
+++ b/packages/@pollyjs/adapter/src/-private/interceptor.js
@@ -1,0 +1,35 @@
+const ABORT = Symbol();
+const PASSTHROUGH = Symbol();
+
+function setDefaults(interceptor) {
+  interceptor[ABORT] = false;
+  interceptor[PASSTHROUGH] = false;
+}
+
+export default class Interceptor {
+  constructor() {
+    setDefaults(this);
+  }
+
+  abort() {
+    setDefaults(this);
+    this[ABORT] = true;
+  }
+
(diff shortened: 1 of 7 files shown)


packages/@pollyjs/adapter/src/-private/interceptor.js
```diff
@@ -0,0 +1,35 @@
...
+  interceptor[ABORT] = false;
+  interceptor[PASSTHROUGH] = false;
+}
+
+export default class Interceptor {
+  constructor() {
+    setDefaults(this);
+  }
+
+  abort() {
```
jasonmit: Might confuse people that this is aborting requests, i.e., https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

offirgolan:
What are your thoughts on `eject` instead?

PR #58, opened 2018-07-08
refactor: Normalized Response + fast-json-stable-stringify

offirgolan:


packages/@pollyjs/adapter/src/index.js
```diff
@@ -144,7 +144,19 @@ export default class Adapter {
       await this.timeout(pollyRequest, recordingEntry);
       pollyRequest.action = ACTIONS.REPLAY;
 
-      return this.onReplay(pollyRequest, recordingEntry);
+      const { status, statusText, headers, content } = recordingEntry.response;
+      const normalizedResponse = {
```
offirgolan: Create a normalized response for all adapter to use. It would be nice if we could use the Response class here from core.

packages/@pollyjs/core/package.json
```diff
@@ -52,8 +52,8 @@
     "@pollyjs/utils": "^0.1.1",
     "@sindresorhus/fnv1a": "^1.0.0",
     "blueimp-md5": "^2.10.0",
+    "fast-json-stable-stringify": "^2.0.0",
```
offirgolan: This is used by a lot of our dependencies. It has no dependents and is proven to be faster.

packages/@pollyjs/persister/package.json
```diff
@@ -44,7 +44,9 @@
   "dependencies": {
     "@pollyjs/utils": "^0.1.1",
     "bowser": "^1.9.3",
+    "fast-json-stable-stringify": "^2.0.0",
     "lodash-es": "^4.17.10",
+    "set-cookie-parser": "^2.2.0",
```
offirgolan: Parses the `set-cookie` header to create HAR compliant cookie objects.

packages/@pollyjs/persister/src/index.js
```diff
@@ -135,6 +136,10 @@ export default class Persister {
     );
   }
 
+  stringify() {
```
offirgolan: Added a utility method here to packages extending this base class wont have to have `fast-json-stable-stringify` as a dependency. 

Not the biggest fan of this but its needed. @jasonmit lemme know if you have a better suggestion.

jasonmit: Another option is @pollyjs/utils

PR #62, opened 2018-07-12
feat: Improved adapter and persister registration

offirgolan:
__Adapters__

```js
import { XHRAdapter, FetchAdapter } from '@pollyjs/core';

// Register the xhr adapter so its accessible by all future polly instances
Polly.register(XHRAdapter);

polly.configure({
  adapters: ['xhr', FetchAdapter]
});
```

__Persister__

```js
import { LocalStoragePersister, RESTPersister } from '@pollyjs/core';

// Register the local-storage persister so its accessible by all future polly instances
Polly.register(LocalStoragePersister);

polly.configure({
  persister: 'local-storage'
});

polly.configure({
  persister: RESTPersister
});
```

### DIFF
--- a/packages/@pollyjs/core/src/-private/container.js
+++ b/packages/@pollyjs/core/src/-private/container.js
@@ -1,5 +1,9 @@
 import { assert } from '@pollyjs/utils';
 
+function keyFor(Factory) {
+  return `${Factory.type}:${Factory.name}`;
+}
+
 export default class Container {
   constructor() {
     this._registry = new Map();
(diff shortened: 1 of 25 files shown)


packages/@pollyjs/core/src/-private/container.js
```diff
@@ -1,17 +1,78 @@
...
+  }
+
+  /**
+   * Check if a factory has been registered by the
+   * given key (e.g. `adapter:fetch`)
+   *
+   * @param {String} key
+   * @returns {Boolean}
+   */
+  has(key) {
```
jasonmit: Should this also take a fn?

packages/@pollyjs/core/src/polly.js
```diff
@@ -119,44 +121,32 @@ export default class Polly {
...
 
-    // Handle Adapters
-    this.config.adapters.forEach(adapter => {
-      let adapterName = adapter;
+    /* Handle Adapters */
 
-      if (isArray(adapterName)) {
-        const [name, AdapterType] = adapterName;
+    // Disconnect from all current adapters
+    this.disconnect();
```
jasonmit: Are there tests asserting disconnect on configure (quick grep didn't find anything)?  Would be useful if there isn't.

packages/@pollyjs/core/src/-private/container.js
jasonmit: Might be useful for it to take a factory also (but not necessary).

packages/@pollyjs/core/src/polly.js
offirgolan: Done.

PR #66, opened 2018-07-16
feat: Fetch adapter support for `context` provided via adapterOptions

### LINKED ISSUE #56: Add ability to configure window in fetch adapter.
In E2E frameworks like [cypress](https://www.cypress.io/) `fetch` is attached to an injected window object like this:

```js
cy.visit('my/pathname/', {
  onBeforeLoad: win => { ... }
})
```

Currently the fetch adapter expects `fetch` to be attached to `self` but in this case it's not so it would be nice if I could supply the window object to the fetch adapter or Polly constructor. E.g.
```js
cy.visit('my/pathname/', {
  onBeforeLoad: win => {
    const polly = new Polly('my-polly-recording', {
      adapters: [['fetch', FetchAdapter.create({ window: win })]]
    });
  }
})
```

jasonmit:
I'll update the docs also.

I can't  apply the same treatment to the XHR adapter since `nise` doesn't enable providing a context object.  I think we spoke about this already, but mentioning again.

Resolves #56

### DIFF
--- a/packages/@pollyjs/adapter-fetch/src/index.js
+++ b/packages/@pollyjs/adapter-fetch/src/index.js
@@ -1,23 +1,32 @@
 import Adapter from '@pollyjs/adapter';
 import { Fetch as FetchUtils } from '@pollyjs/utils';
 
-const nativeFetch = global.fetch;
 const { defineProperty } = Object;
+const STUB_META = Symbol();
 
 export default class FetchAdapter extends Adapter {
   static get name() {
     return 'fetch';
   }
 
+  get defaultOptions() {
+    return {
+      context: global
+    };
+  }
+
   onConnect() {
-    this.assert('Fetch global not found.', nativeFetch);
+    const { context } = this.options;
+
+    this.assert('Fetch global not found.', !!(context && context.fetch));
     this.assert(
       'Running concurrent fetch adapters is unsupported, stop any running Polly instances.',
-      global.fetch === nativeFetch
+      !(STUB_META in context.fetch)
     );
 
-    this.native = nativeFetch;
-    global.fetch = (url, options = {}) =>
+    this.native = context.fetch;
+
+    context.fetch = (url, options = {}) =>
       this.handleRequest({
         url,
         method: options.method || 'GET',
@@ -25,10 +34,13 @@ export default class FetchAdapter extends Adapter {
         body: options.body,
         requestArguments: [url, options]
       });
+
+    defineProperty(context.fetch, STUB_META, { value: true });
   }
 
   onDisconnect() {
-    global.fetch = nativeFetch;
+    this.options.context.fetch = this.native;
+    this.native = null;
   }
 
   async onRecord(pollyRequest) {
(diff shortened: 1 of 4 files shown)


packages/@pollyjs/adapter-fetch/tests/integration/adapter-test.js
```diff
@@ -21,3 +21,36 @@ describe('Integration | Fetch Adapter', function() {
...
+
+describe('Integration | Fetch Adapter | Concurrency', function() {
+  it('should prevent concurrent fetch adapter instances', function() {
+    const one = new Polly('one');
+    const two = new Polly('two');
+
+    one.connectTo(FetchAdapter);
+
+    expect(function() {
+      two.connectTo(FetchAdapter);
```
jasonmit: `connectTo` vs providing the adapter via `configure` so that the constructor assignment happens and we can do cleanup/`stop()`.

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -1,34 +1,46 @@
...
+    return {
+      context: global
+    };
+  }
+
   onConnect() {
-    this.assert('Fetch global not found.', nativeFetch);
+    const { context } = this.options;
+
+    this.assert('Fetch global not found.', context.fetch);
```
offirgolan: `!!(context && context.fetch)` 

Assert only checks for false or undefined. Might wanna open that up to any falsey value.

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -1,34 +1,46 @@
...
+    context.fetch = (url, options = {}) =>
       this.handleRequest({
         url,
         method: options.method || 'GET',
         headers: FetchUtils.serializeHeaders(options.headers),
         body: options.body,
         requestArguments: [url, options]
       });
+
+    defineProperty(context.fetch, STUB_META, { value: {} });
```
offirgolan: Why a meta object vs a Boolean?

packages/@pollyjs/adapter-fetch/tests/integration/adapter-test.js
```diff
@@ -21,3 +21,36 @@ describe('Integration | Fetch Adapter', function() {
   adapterTests();
   adapterBrowserTests();
 });
+
+describe('Integration | Fetch Adapter | Concurrency', function() {
```
offirgolan: Can you also add a test against a null context and one without fetch?

offirgolan:
> I can't apply the same treatment to the XHR adapter 

I just meant the logic around checking the STUB_META instead of checking against the captured native class in the module. That way, we can still wrap a pre modified XHR.

packages/@pollyjs/adapter-fetch/tests/integration/adapter-test.js
jasonmit: Yup need to add tests in general around context.  Will finish up tonight + docs

jasonmit:
> I just meant the logic around checking the STUB_META instead of checking against the captured native class in the module. That way, we can still wrap a pre modified XHR.

I'll tackle in a separate PR tomorrow.

PR #67, opened 2018-07-17
fix: Creator cleanup and persister assertion

offirgolan:


tests/recordings/-pollyjs_2160168770/adapter-fetch_162589637/Integration-Fetch-Adapter_1763210894/should-be-able-to-abort-from-an-intercept_68559697/recording.har
```diff
@@ -6,6 +6,7 @@
       "version": "67.0"
     },
     "creator": {
+      "comment": "persister:rest",
```
offirgolan: Not sure if we want this or something like `via the rest persister`...

jasonmit: this is fine.  just needs some test coverage

PR #70, opened 2018-07-19
test: Add jest smoke tests

offirgolan:
Add XHR & Fetch adapter smoke tests to make sure they behave nicely in JSDom.

### DIFF
--- a/packages/@pollyjs/adapter-fetch/tests/jest/integration/adapter-test.js
+++ b/packages/@pollyjs/adapter-fetch/tests/jest/integration/adapter-test.js
@@ -1,0 +1,41 @@
+import '@pollyjs-tests/helpers/global-fetch';
+
+import { Polly } from '@pollyjs/core';
+import FSPersister from '@pollyjs/persister-fs';
+import FetchAdapter from '../../../src';
+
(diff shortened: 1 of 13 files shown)


packages/@pollyjs/adapter-fetch/tests/jest/integration/adapter-test.js
```diff
@@ -0,0 +1,41 @@
...
+describe('Integration | Fetch Adapter | Jest', function() {
+  let polly;
+
+  beforeEach(() => {
+    polly = new Polly('Integration | Fetch Adapter | Jest', {
+      recordFailedRequests: true,
+      adapters: [FetchAdapter],
+      persister: FSPersister,
+      persisterOptions: {
+        fs: { recordingsDir: 'tests/recordings' }
```
jasonmit: We could supply the fs persister a mock fs library to prevent it from ever writing to disk in our tests (avoids the `delete()` but might not be called if an exception was thrown).

offirgolan: We could do that in a different PR if we need to. I sort of like this more since we are testing the full lifecycle / the entire persister API.

PR #75, opened 2018-07-23
feat: Wait for all handled requests to resolve via `.flush()`

### LINKED ISSUE #71: How to wait for all the responses to finish?
Is there anything like `fetchMock.fush()` ?
http://www.wheresrhys.co.uk/fetch-mock/api#flush
This is how they do it:
```
FetchMock.flush = function() {
	return Promise.all(this._holdingPromises);
};
```
```
FetchMock.fetchHandler = function(url, opts) {
	const response = this.executeRouter(url, opts);

	[...]

	// this is used to power the .flush() method
	let done;
	this._holdingPromises.push(new this.config.Promise(res => (done = res)));

	// wrapped in this promise to make sure we respect custom Promise
	// constructors defined by the user
	return new this.config.Promise((res, rej) => {
		this.generateResponse(response, url, opts)
			.then(res, rej)
			.then(done, done);
	});
};
```

offirgolan:
Resolves #71.


packages/@pollyjs/core/src/polly.js
```diff
@@ -234,6 +234,14 @@ export default class Polly {
     }
   }
 
+  async flush() {
+    const NOOP = () => {};
```
jasonmit: Whats the roll of the noops here

packages/@pollyjs/core/src/polly.js
```diff
@@ -234,6 +234,14 @@ export default class Polly {
...
+
+    await Promise.all(
+      this._requests.map(r => Promise.resolve(r.promise).then(NOOP, NOOP))
```
jasonmit: Whats the role of the noops here

packages/@pollyjs/utils/src/utils/deferred-promise.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Create a deferred promise with `resolve` and `reject` methods.
+ *
+ * @returns {Promise}
+ */
+export default class DeferredPromise extends Promise {
```
jasonmit: Could be part of core since it's only used there

tests/integration/adapter-tests.js
```diff
@@ -99,4 +99,30 @@ export default function adapterTests() {
...
+      res.sendStatus(200);
+    });
+
+    this.fetchRecord();
+    this.fetchRecord();
+    this.fetchRecord();
+
+    // Wait for all requests to be made
+    // (Puppeteer adapter takes some time to make the actual request)
+    while (requests.length !== 3) {
```
jasonmit: Is this only noticed within the puppeteer adapter?  Worry we could mask potential future bugs by doing this always in all the adapter tests.

offirgolan: > Is this only noticed within the puppeteer adapter?

Yes.

We use these tests in many places for both node and browser. I'm not really sure whats a better way to do this while making sure all adapters are tested for.

offirgolan: > I wonder also if this could be mitigated by resolving the flush promise in the next frame.

What do you mean by this?

jasonmit: > We use these tests in many places for both node and browser. I'm not really sure whats a better way to do this while making sure all adapters are tested for.

Skip the test in puppeteer.  Then create a test with this while loop within puppeteer's test suite so that we don't mask future bugs on the other adapters. 

> I wonder also if this could be mitigated by resolving the flush promise in the next frame.

Lets chat, I'd like to know more about what is going on here.  Specifically "takes some time".  Do we know how much time?  Does this while loop loop more than once?

packages/@pollyjs/adapter-puppeteer/tests/integration/adapter-test.js
```diff
@@ -64,4 +64,40 @@ describe('Integration | Puppeteer Adapter', function() {
...
+      })
+      .on('request', req => requests.push(req));
+
+    this.page.on('requestfinished', () => resolved.push(i++));
+
+    this.fetchRecord();
+    this.fetchRecord();
+    this.fetchRecord();
+
+    // Since it takes time for Puppeteer to execute the request in the browser's
```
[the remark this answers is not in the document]
offirgolan: As discussed, there are no currently known edge cases. This is a test nuance that only relates to this test scenario.

PR #78, opened 2018-07-24
fix: Improve support for relative URLs

### LINKED ISSUE #76: Host is stripped from protocol-relative urls
**How to reproduce:**
1. Have app on `http://my-app.dev`
1. Make request to protocol-relative url `//api.dev/posts/1`

**Expected behavior**
Polly proxies this request to `api.dev` and writes response.

**Actual behavior**
Polly strips host from `//api.dev/posts/1` making it into `/posts/1` and then trying to send request to `http://my-app.dev/posts/1` instead of original url.

I found two places in code where url is modified. Here is one which caused desribed behavior:
https://github.com/Netflix/pollyjs/blob/73026d446cf207ab814db5fb653bdfee40cfe682/packages/%40pollyjs/core/src/-private/request.js#L53-L60

Seems like [is-absolute-url](https://www.npmjs.com/package/is-absolute-url) check here can be swapped with custom implementation supporting protocol-relative urls

offirgolan:
- Create a shared parseUrl utility that will make sure to match the provided URL string to a url-parse instance
- Support for protocol-relative URLs (Resolves #76)

### DIFF
--- a/packages/@pollyjs/core/tests/unit/utils/parse-url-test.js
+++ b/packages/@pollyjs/core/tests/unit/utils/parse-url-test.js
@@ -13,4 +13,12 @@ describe('Unit | Utils | parseUrl', function() {
       'http://netflix.com/movies/1?sort=title&dir=asc'
     ].forEach(url => expect(parseUrl(url).href).to.equal(url));
   });
+
+  it('should passthrough arguments to url-parse', function() {
+    // Passing true tells url-parse to transform the querystring into an object
+    expect(parseUrl('/movies/1?sort=title&dir=asc', true).query).to.deep.equal({
+      sort: 'title',
+      dir: 'asc'
+    });
+  });
 });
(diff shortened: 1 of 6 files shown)


packages/@pollyjs/core/tests/unit/utils/parse-url-test.js
```diff
@@ -0,0 +1,16 @@
...
+    expect(parseUrl).to.be.a('function');
+  });
+
+  it('should exactly match passed urls', function() {
+    [
+      '/movies/1',
+      '//netflix.com/movies/1',
+      'http://netflix.com/movies/1',
+      'http://netflix.com/movies/1?sort=title&dir=asc'
+    ].forEach(url => expect(parseUrl(url).href).to.equal(url));
```
jasonmit: Can you add a test for the `true` argument to `parseUrl`.  Wasn't really sure how it was being used.

PR #80, opened 2018-07-30
feat: Server level configuration

offirgolan:
### intercept

Intercept can now be registered on the middleware level. If multiple intercept handlers have been
registered, each handler will be called in the order in which it was registered.

__Example__

```js
server
  .any('/session')
  .intercept((req, res) => res.sendStatus(200));
```

### passthrough

Passthrough can now be registered on the middleware level. It also now accepts a boolean which allows the user to explicitly enable or disable it.

| Param | Type | Description |
|  ---  | ---  |     ---     |
| passthrough | `boolean` | Enable or disable the passthrough. Defaults to `true` |

__Example__

```js
server
  .any('/session')
  .passthrough();

server
  .get('/session/1')
  .passthrough(false);
```

### recordingName <sup>_new_</sup>

Override the recording name for the given route. This allows for grouping common
requests to share a single recording which can drastically help de-clutter test
recordings.

For example, if your tests always make a `/users` or `/session` call, instead of
having each of those requests be recorded for every single test, you can use
this to create a common recording file for them.

| Param | Type | Description |
|  ---  | ---  |     ---     |
| recordingName | `String` | Name of the [recording](api#recordingName) to store the recordings under. |

__Example__

```js
server
  .any('/session')
  .recordingName('User Session');

server
  .get('/users/:id')
  .recordingName('User Data');

server
  .get('/users/1')
  .recordingName(); /* Fallback to the polly instance's recording name */
```

### configure <sup>_new_</sup>

Override configuration options for the given route. All matching middleware and route level configs are merged together and the overrides are applied to the current
Polly instance's config.

> The following options are not supported to be overriden via the server API:
`mode`, `adapters`, `adapterOptions`, `persister`, `persisterOptions`

| Param | Type | Description |
|  ---  | ---  |     ---     |
| config | `Object` | [Configuration](configuration) object |

__Example__

```js
server
  .any('/session')
  .configure({ recordFailedRequests: true });

server
  .get('/users/:id')
  .configure({ timing: Timing.relative(3.0) });

server
  .get('/users/1')
  .configure({ logging: true });
```


jasonmit:
I'm worried `req.configure()` will be confused with `polly.configure()` as I imagine we can only support a subset of options here.

It's unclear when one would use the request-level configuration over the server/route level.  Should we simplify and only support one for now until there is a clear use-case?

packages/@pollyjs/core/tests/browser/integration/server-test.js
```diff
@@ -4,6 +4,63 @@ import FetchAdapter from '@pollyjs/adapter-fetch';
...
+    [
+      'mode',
+      'adapters',
+      'adapterOptions',
+      'persister',
+      'persisterOptions'
+    ].forEach(key =>
+      expect(() => server.any().configure({ [key]: 'foo' })).to.throw(
+        Error,
+        /Invalid configuration option/
```
jasonmit: Error could include these options not being allowed to be configure through the new config API vs invalid configuration options since they're technically valid.

packages/@pollyjs/core/tests/unit/utils/timing-test.js
```diff
@@ -3,7 +3,7 @@ import Timing from '../../../src/utils/timing';
 function fixedTest(ms) {
   it(`should handle ${ms}ms`, async function() {
     // Fail the test if it exceeds ms + 10ms buffer
-    this.timeout(ms + 10);
+    this.timeout(ms + 50);
```
jasonmit: thanks!

packages/@pollyjs/core/tests/browser/integration/server-test.js
offirgolan: The entire error message is:

```js
`Invalid configuration option provided. The "${key}" option cannot be overridden.`
```

PR #88, opened 2018-08-08
fix(persister): Handle concurrent find requests

offirgolan:


packages/@pollyjs/persister/src/index.js
```diff
@@ -116,31 +116,40 @@ export default class Persister {
...
+
+        if (recording) {
+          this.assert(
+            `Recording with id '${recordingId}' is invalid. Please delete the recording so a new one can be created.`,
+            recording.log && recording.log.creator.name === CREATOR_NAME
+          );
+
+          return recording;
+        } else {
+          cache.delete(recordingId);
```
jasonmit: Curious why the delete here?

packages/@pollyjs/persister/src/index.js
```diff
@@ -116,31 +116,40 @@ export default class Persister {
   }
 
   async find(recordingId) {
-    if (this.cache.has(recordingId)) {
-      return this.cache.get(recordingId);
-    }
+    const { _cache: cache } = this;
+
+    if (!cache.has(recordingId)) {
+      const findRecording = async () => {
```
jasonmit: Might be worth naming this anonymous fn for a better callstack

PR #90, opened 2018-08-09
fix(adapter-puppeteer): Do not intercept CORS preflight requests

### LINKED ISSUE #87: Turning on Polly makes puppeteer do hundreds of requests
I am using jest + puppeteer to run tests. I am trying to use polly to record and serve requests to our API  when running integration tests.

What happens when i turn on Polly to record our requests is that it starts running the same request for hundreds of times.

The packages I am using: PuppeteerAdapter, FSPersister.
When I turn off Polly the request completes successfully.

I recorded a video of the chromium's console:
https://gfycat.com/EssentialShadowyAnura

Here are the versions I am using:

```
    "@pollyjs/adapter-puppeteer": "^1.1.0",
    "@pollyjs/core": "^1.1.0",
    "@pollyjs/persister-fs": "^1.0.1",
```

and the configs:

**Server**
```
{
      recordingsDir: path.join(__dirname, 'recordings'),
      quiet: false,
      port: 3001,
      host: 'localhost'
}
```

**Polly**

```
{
      adapters: ['puppeteer'],
      adapterOptions: {
        puppeteer: { page }
      },
      persister: 'fs',
      persisterOptions: {
        host: 'http://localhost:3001'
      },
      logging: true
}
```

Is there something I am missing in the config? Or in general?

offirgolan:
Resolves #87.

### DIFF
--- a/packages/@pollyjs/adapter-puppeteer/src/index.js
+++ b/packages/@pollyjs/adapter-puppeteer/src/index.js
@@ -1,10 +1,11 @@
 import Adapter from '@pollyjs/adapter';
+import { URL } from '@pollyjs/utils';
 
 const LISTENERS = Symbol();
 const POLLY_REQUEST = Symbol();
 const PASSTHROUGH_PROMISE = Symbol();
 const PASSTHROUGH_PROMISES = Symbol();
-const PASSTHROUGH_REQ_ID_HEADER = 'x-pollyjs-passthrough-request-id';
+const PASSTHROUGH_REQ_ID_QP = 'pollyjs_passthrough_req_id';
 
 export default class PuppeteerAdapter extends Adapter {
   static get name() {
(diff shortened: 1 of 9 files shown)


jasonmit:
CORs test is failing in CI

packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -191,47 +183,29 @@ export default class PuppeteerAdapter extends Adapter {
...
+          parsedUrl.toString(),
+          { method, headers, body }
+        );
       });
-      let responseBody;
-
-      try {
-        responseBody = await response.text();
-      } catch (error) {
-        /*
```
offirgolan: This is no longer needed since im resolving the promise on `requestfinished` instead of on `response`. That's where it should have been resolving in the first place.

packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -1,10 +1,11 @@
 import Adapter from '@pollyjs/adapter';
+import { URL } from '@pollyjs/utils';
```
offirgolan: Using `url-parse` instead of node's `URL` as it is a consistent API across node versions and the browser.

PR #94, opened 2018-08-14
chore: Upgrading dependencies

jasonmit:


package.json
```diff
@@ -41,41 +41,41 @@
...
-    "eslint-plugin-prettier": "^2.6.0",
+    "eslint-plugin-prettier": "^2.6.2",
     "formdata-polyfill": "^3.0.11",
     "gh-pages": "^1.2.0",
     "har-validator": "^5.1.0",
     "husky": "^0.14.3",
-    "jest": "^23.4.1",
+    "jest": "^23.5.0",
     "jest-tap-reporter": "^1.9.0",
     "lerna": "^2.11.0",
```
jasonmit: lerna cannot be bumped yet to 3.0.0 since commitlint lerna scopes currently doesn't work with it.  Someone has a PR opened to address it.

PR #101, opened 2018-08-20
fix: Allow 204 responses without a body

cibernox:
It turns out that 204 responses using fetch attempt to call `new Response`
with an empty string as a body, which is illegal and blows hard. The body
must be null.


jasonmit:
I'll add a test or two before versioning.  Thanks!

cibernox:
Yeah, I tried for 20m to find a good way to test it but I wasn't very sure how to do it.

PR #102, opened 2018-08-21
chore: Adds tests to cover #101

jasonmit:
@cibernox I believe this will cover your change

### DIFF
--- a/tests/integration/adapter-tests.js
+++ b/tests/integration/adapter-tests.js
@@ -26,6 +26,13 @@ export default function adapterTests() {
     expect(res.status).to.equal(404);
   });
 
+  it('should properly handle 204 status code response', async function() {
+    const noContentResponse = await this.relativeFetch('/echo?status=204');
+
+    expect(noContentResponse.status).to.equal(204);
+    expect((await noContentResponse.text())).to.equal('');
+  });
+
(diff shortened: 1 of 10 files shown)


tests/integration/adapter-tests.js
```diff
@@ -26,6 +26,19 @@ export default function adapterTests() {
     expect(res.status).to.equal(404);
   });
 
+  it('should handle 204s with empty response body', async function() {
+    let res = await this.fetchRecord({
+      method: 'POST',
+      body: `''`
```
jasonmit: Actually in doing this, I think the bug is elsewhere.  I shouldn't need to pass an empty string here to mimic the bug.

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -89,9 +89,7 @@ export default class FetchAdapter extends Adapter {
     const { absoluteUrl, response } = pollyRequest;
     const { statusCode } = response;
 
-    const responseBody =
```
jasonmit: Moved this logic into the polly response object since we create Response instances with that object in a few places.

cibernox:
Makes sense. Can you publish a new release?

PR #104, opened 2018-09-06
Babel 7 support

jasonmit:


packages/@pollyjs/core/tests/browser/unit/utils/serialize-request-body-test.js
```diff
@@ -23,24 +23,31 @@ describe('Unit | Utils | serializeRequestBody', function() {
   });
 
   it('should handle files', async function() {
-    expect(await serializeRequestBody(new File(['file'], 'file.txt'))).to.equal(
-      `data:;base64,${btoa('file')}`
-    );
+    expect(
+      await serializeRequestBody(
+        new File(['file'], 'file.txt', {
```
jasonmit: > readAsDataURL function is supported, but the mediatype in Chrome depends on entry name extension, mediatype in IE is always empty (which is the same as text-plain according the specification), the mediatype in Firefox is always application/octet-stream. For example, if the content is abcdefg then Firefox returns data:application/octet-stream;base64,YWJjZGVmZw==, IE returns data:;base64,YWJjZGVmZw==, Chrome returns data:<mediatype depending on extension of entry name>;base64,YWJjZGVmZw==.

Seems how the type is negotiated when not supplied is a browser quirk.  In Chrome 68 if the type of `''` it would not default to `application/octet-stream` - now it does... like Firefox.

package.json
```diff
@@ -67,17 +68,20 @@
...
-    "rollup-plugin-uglify": "^3.0.0",
+    "rollup-plugin-node-globals": "^1.3.0",
+    "rollup-plugin-node-resolve": "^3.4.0",
+    "rollup-plugin-terser": "^2.0.2",
     "tap-merge": "^0.3.1",
-    "testem": "^2.9.2"
+    "testem": "^2.10.0"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.0.0"
```
jasonmit: intentional for polly to work in node 6

offirgolan: @jasonmit I think this should go in devDependencies

PR #106, opened 2018-09-07
fix: Browser (UMD) build now bundles corejs

jasonmit:


build-scripts/rollup.node.config.js
```diff
@@ -16,19 +16,19 @@ export default function createNodeConfig(options = {}) {
       external,
       plugins: [
         json(),
-        resolve(),
+        resolve({ preferBuiltins: true }),
```
jasonmit: The default, but suppressing warnings when you explicitly set it.

PR #108, opened 2018-09-13
feat(persister): Add `keepUnusedRequests` config option

offirgolan:
Currently by default, all new requests are appended to the currently existing recording HAR file. Although this works well (especially when using Polly for an offline only tool), it creates bloated recording files with unused requests. This PR handles the more generic case where a recording file should only contain the requests that have been captured by the running Polly instance. 

### keepUnusedRequests _<sup>new</sup>_

 _Type_: `Boolean`
_Default_: `false`

When disabled, requests that have not been captured by the running Polly instance will be removed. This ensures that a recording will only contain the requests that were made during the lifespan of the Polly instance. When enabled, new requests will be appended to the recording file.

 __Example__

 ```js
polly.configure({
  persisterOptions: {
    keepUnusedRequests: true
  }
});
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -223,7 +223,7 @@ _Default_: `false`
 
 When disabled, requests that have not been captured by the running Polly
 instance will be removed. This ensures that a recording will only contain the
-necessary requests that were made during the lifespan of the Polly instance.
+requests that were made during the lifespan of the Polly instance.
 When enabled, new requests will be appended to the recording file.
 
 __Example__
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/configuration.md [hunk 1]
```diff
@@ -216,6 +216,26 @@ polly.configure({
 });
 ```
 
+### keepUnusedRequests
+
+_Type_: `Boolean`
+_Default_: `false`
+
+When disabled, requests that have not been captured by the running Polly
+instance will be removed. This ensures that a recording will only contain the
```
jasonmit: "will be removed from any previous recording."

PR #110, opened 2018-09-15
fix: Puppeteer CORS request matching

offirgolan:
Preflight requests should not be matched to a pollyRequest since Polly should never actually be aware of such a request.


packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -231,6 +233,7 @@ export default class PuppeteerAdapter extends Adapter {
         await response.text()
       );
     } catch (error) {
+      console.error(error);
```
offirgolan: Print out the error so its displayed in the test runners.

Issue #120, opened 2018-10-25
Node HTTP Request Adapter


offirgolan:
Create a low level Node HTTP Request adapter so we can fully support Node use cases.

yasinuslu:
Hi @offirgolan, i would like to work on this

offirgolan:
@yasinuslu thats awesome! PRs are always welcomed. Feel free to just go for it or open an RFC if you'd like to discuss implementation before 😄

yasinuslu:
Thanks @offirgolan, i'll start working on it and see if i need to open an RFC or if i can just go for an implementation

PR #128, opened 2018-11-05
feat: Node HTTP Request adapter

yasinuslu:
:warning: This PR is still a WIP

Hello everyone,

This will hopefully introduce HTTP Request adapter mentioned in #120 
I wanted to request your comments as i implement the rest.


offirgolan:
@yasinuslu let me know when you're ready for review 😄. I've fixed the hanging test case so feel free to rebase from master.

yasinuslu:
@offirgolan adapter's tests are passing in nodejs 10. I'm investigating why they fail at nodejs 6 and 8. Other than that it's ready for review.

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      const resBuffer = [];
+
+      stream.on('data', chunk => {
+        resBuffer.push(chunk);
+      });
+
+      stream.on('end', () => {
+        const data = Buffer.concat(resBuffer).toString('utf8');
```
offirgolan: Do you see any issues with storing the data as `utf8`? Will there be an issues with replaying content with other encodings?

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,198 @@
...
+          stream = res.pipe(zlib.createUnzip());
+          delete res.headers['content-encoding'];
+          break;
+        default:
+          stream = res;
+      }
+
+      const resBuffer = [];
+
+      stream.on('data', chunk => {
```
offirgolan: `stream.on('data', chunk => resBuffer.push(chunk) });`

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,198 @@
...
+const LISTENERS = Symbol();
+
+export default class TransportWrapper {
+  constructor(transport, { adapter }) {
+    this.adapter = adapter;
+    this.transport = transport;
+  }
+
+  patch() {
+    if (!this.transport[NATIVE_REQUEST]) {
```
offirgolan: We should throw here if we've already patched the transport.

packages/@pollyjs/adapter-http/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@pollyjs/adapter-http",
```
offirgolan: We should rename this to `adapter-node-http` just to denote that this a node only adapter.

packages/@pollyjs/adapter-http/src/index.js
```diff
@@ -0,0 +1,61 @@
+import Adapter from '@pollyjs/adapter';
+
+import HttpWrapper from './utils/http-wrapper';
+
+export default class HttpAdapter extends Adapter {
+  static get name() {
+    return 'http';
```
offirgolan: `note-http`

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,198 @@
...
+       */
+
+      const req = transport[NATIVE_REQUEST].call(transport, ...args);
+
+      const chunks = [];
+
+      const originalWrite = req.write;
+
+      req.write = (chunk, encoding, callback) => {
+        // TODO : handle encoding
```
offirgolan: Is this still a todo?

packages/@pollyjs/adapter-http/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "description": "Node http adapter for @pollyjs",
+  "main": "dist/cjs/pollyjs-adapter-http.js",
+  "module": "dist/es/pollyjs-adapter-http.js",
+  "files": [
+    "src",
+    "dist"
+  ],
+  "repository": "https://github.com/netflix/pollyjs/tree/master/packages/@pollyjs/adapter-http",
+  "license": "Apache-2.0",
+  "contributors": [
```
jasonmit: Your name goes here :)

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
yasinuslu: I haven't tried with a binary content but it'll probably fail right now. Maybe we can store binary content as `base64` ?

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
yasinuslu: I'm still not sure how to handle binary content, i left it there to not forget it :)

packages/@pollyjs/adapter-http/src/utils/http-wrapper.js
```diff
@@ -0,0 +1,21 @@
+import http from 'http';
+import https from 'https';
+
+import TransportWrapper from './transport-wrapper';
+
+export default class HttpWrapper {
+  constructor({ adapter }) {
+    this.transports = [
+      new TransportWrapper(http, { adapter }),
```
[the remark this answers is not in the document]
yasinuslu: Something like this will do ?
```js
get defaultOptions() {
  return {
    transports: ['http', 'https']
  };
}
```

packages/@pollyjs/adapter-http/package.json
yasinuslu: Thanks :)

packages/@pollyjs/adapter-http/src/utils/http-wrapper.js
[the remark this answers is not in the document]
yasinuslu: Done in 47ab609959ee4d4b12d2553fbe1288e1fb53ddc1

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
yasinuslu: We're already discussing this in other discussion, i can remove this TODO comment now if you want ?

packages/@pollyjs/adapter-node-http/src/index.js
```diff
@@ -0,0 +1,69 @@
...
+import HttpWrapper from './utils/http-wrapper';
+
+export default class HttpAdapter extends Adapter {
+  static get name() {
+    return 'node-http';
+  }
+
+  get defaultOptions() {
+    return {
+      transports: {
```
offirgolan: For consistency with the other adapters, I'd prefer this to be `['http', 'https']`.

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  isPatched() {
+    return !!this.transport[NATIVE_REQUEST];
+  }
+
+  patch() {
+    // make sure it's not already patched
+    this.adapter.assert(
+      `Transport ${this.name} can not be patched multiple times`,
```
offirgolan: `The ${this.name} transport has already been patched, please stop any running Polly instances.`

packages/@pollyjs/adapter-node-http/src/utils/http-wrapper.js
```diff
@@ -0,0 +1,35 @@
...
+import https from 'https';
+
+import TransportWrapper from './transport-wrapper';
+
+export default class HttpWrapper {
+  constructor({ adapter, options }) {
+    this.transports = [];
+
+    adapter.assert(
+      'There should be at least one transport enabled',
```
offirgolan: `Invalid transports provided. At least one supported transport must be specified.`

offirgolan: This can also be moved to the adapter's onConnect. That way you can assume that the HttpWrapper always receives valid options. Totally up to you though 😄

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
offirgolan: Im ok with storing binary data as `base64`, I do have a few questions/concerns:

1. Do we have a way of determining whether we can decode the buffer content in utf-8 or fallback to base64? Can we use response headers or maybe something like [is-utf8](https://www.npmjs.com/package/is-utf8)?
2. When we respond with the base64 content, do we need to provide the original encoding to the buffer? If so, should we persist the encoding in the recordings (I can help with that if necessary)?


@yasinuslu can you also add a test case for this as well?

offirgolan: Follow up question:

Do we support for sending binary data in the request payload? We should add a test case for this as well. If not, I can help on the @pollyjs/core side to try to get this working.

packages/@pollyjs/adapter-node-http/src/utils/http-wrapper.js
yasinuslu: Done in 25589639be340c56e0fcd56d85d13c36585fde2f

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: Done in a3f7248831b7540af93d7a4542ec7579a863a5c2

packages/@pollyjs/adapter-node-http/src/index.js
yasinuslu: Done in 25589639be340c56e0fcd56d85d13c36585fde2f

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,198 @@
...
+
+export default class TransportWrapper {
+  constructor(transport, { adapter }) {
+    this.adapter = adapter;
+    this.transport = transport;
+  }
+
+  patch() {
+    if (!this.transport[NATIVE_REQUEST]) {
+      this.transport[NATIVE_REQUEST] = this.transport.request;
```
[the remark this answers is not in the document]
yasinuslu: Done in d8acd6d0ffcc435025800af55e533028d030ab0b

packages/@pollyjs/adapter-http/src/utils/transport-wrapper.js
yasinuslu: @offirgolan Of course, i'll add a test case for this. 

I don't have the answers you're looking for right now, i'll get back to you after some experimenting 😄

yasinuslu: I did some testing on this and i think it's best if we store every response body as `base64`. 
So in response to your questions if we store `base64`:
1. We won't need to determine which encoding to use.
2. We will respond with a buffer with original bytes because we respond with `Buffer.from(storedBody, 'base64')` we can construct same bytes whether they were binary or utf8.
3. We can probably support binary data in the request payload if we also store it as `base64`.

I successfully replayed both utf8 and binary data in my local, i'll add test cases for both binary request payload and binary responses. Maybe we can add a new field in the recordings that indicates if the stored body is utf8 encoded or base64.

offirgolan: I did some investigation on my end as well. Looking at nock's implementation, we can do something very similar to determine how to handle both the request and response body:

https://github.com/nock/nock/blob/master/lib/recorder.js#L49-L97

yasinuslu: Thanks, that seems like a nice library. I'll start implementing something similar than.

timpur:
@yasinuslu , any idea when this will be ready ? Kinda need it now :P

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,209 @@
...
+        const body = chunks.join();
+        const [hostname, port = 80] = host.split(':');
+
+        const parsedUrl = new URL('');
+
+        parsedUrl.set('protocol', req.agent.protocol);
+        parsedUrl.set('pathname', path);
+        parsedUrl.set('hostname', hostname);
+        if (port !== 80) {
+          parsedUrl.set('port', port);
```
offirgolan:
```suggestion
          parsedUrl.set('port', port !== 80 ? port : '');
```

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,212 @@
...
+      req.write = (chunk, encoding, callback) => {
+        // TODO : handle encoding
+        chunks.push(chunk);
+
+        originalWrite.call(req, chunk, encoding, callback);
+      };
+
+      // This can be called multiple times with the same request instance
+      // so make sure we only save the original `end` method.
+      req[NATIVE_END] = req[NATIVE_END] || req.end;
```
offirgolan: In Node 6, the transport.request method was getting called multiple times with the same request instance and was going into an infinite loop since `NATIVE_END` was the patched `end` method on line 148.

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,212 @@
...
+          // until we decide what to do in next steps
+          req.removeAllListeners(eventName);
+
+          acc[eventName] = listeners;
+
+          return acc;
+        }, {});
+
+        const headers =
+          typeof req.getHeaders === 'function'
```
offirgolan: `req.getHeaders()` is not a valid method in Node 6. I'm not sure if the changes I've made are correct since the only valid property I could access was `req._headers`. 

@yasinuslu if you could confirm this logic, that would be appreciated 😄

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,212 @@
...
+    // make sure it's not already patched
+    this.adapter.assert(
+      `The ${
+        this.name
+      } transport has already been patched, please stop any running Polly instances`,
+      !this.isPatched()
+    );
+
+    nativeRequestMapping.set(this.transport, this.transport.request);
+    this.transport.request = this.createRequestWrapper();
```
offirgolan: Do we also need to patch `request.get`?

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,212 @@
...
+
+    this.transport.request = nativeRequestMapping.get(this.transport);
+    nativeRequestMapping.delete(this.transport);
+  }
+
+  getHttpResponseData(res) {
+    return new Promise(resolve => {
+      let stream;
+
+      switch (res.headers['content-encoding']) {
```
offirgolan: How do we handle multiple encoding?

```
Content-Encoding: gzip, identity
Content-Encoding: deflate, gzip
```

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: We should cover `http.get` with patching `request`:
```js
function get(url, options, cb) {
  var req = request(url, options, cb);
  req.end();
  return req;
}
```
It didn't seem to change from [2015](https://github.com/nodejs/node/blob/9cd44bb2b683446531306bbcff8739fc3526d02c/lib/http.js#L34) till [now](https://github.com/nodejs/node/blob/7cc3b1eac56d6a5759a298666ad11436396fdf0c/lib/http.js#L44)

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: We currently don't support it. This part needs a second look :)

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: @offirgolan I couldn't find where it gets assigned in `nodejs` source code but i think it will be fine :smile:

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: Good catch, i didn't think we would have a problem like that in here :)

timpur:
If there is anything I can do to help?

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
```diff
@@ -0,0 +1,212 @@
...
+      stream.on('end', () => {
+        const data = Buffer.concat(resBuffer).toString('utf8');
+
+        resolve(data);
+      });
+    });
+  }
+
+  async passthrough(req) {
+    const res = await new Promise(resolve => {
```
offirgolan: I just realized that the way you are handling the passthrough doesnt match how we handle it in the other adapters which I believe will complicate this a bit more. What needs to happen here is the passthrough should create a completely new request and take the headers, url, method, etc from the PollyRequest instance since they can be modified before the request gets sent out. 

Consider the following scenario:

```js
const { server } = polly;

// Modify the authorization request headers before it gets sent out
server.get('/foo').on('request', req => req.setHeader('Authorization', 'abc123'));

// Create my request
const req = http.get('/foo', { headers: { authorization: 'my-auth-token' } });

// After this gets called, the actual passed-through request should contain the modified header
req.end();
```

A good example for how we do this is in the `adapter-xhr`. We create a fake XHR object that never actually makes the request. When `xhr.send()` is called, we create a brand new real XHR instance and send all the data we have present in the PollyRequest.

offirgolan:
@yasinuslu I believe I was able to solve the content encoding and passthrough issues. I've created a [branch](https://github.com/Netflix/pollyjs/compare/content-encoding?expand=1) that I can merge into yours if you're okay with that. I still need to resolve the request body serialization but that wont be too difficult 😄

yasinuslu:
Thanks @offirgolan, i looked into it and i'm ok with it 😄 I can merge it if you want

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -0,0 +1,283 @@
...
+    const { method, headers, body: chunks } = pollyRequest;
+    const nativeRequest = nativeRequestMapping.get(transport);
+    let [url, options] = args;
+
+    /**
+     * args could be (url, options, callback) or (options, callback) depending
+     * on nodejs version. If the `url` arguments is not a string or URL
+     * instance, then use (options, callback).
+     */
+    if (typeof url !== 'string' || !(isObjectLike(url) && url.href)) {
```
yasinuslu: I think there might be a problem here, if url is a string it is used as options:
image
 
and in the next lines it will be spread to `nativeRequest` call:
```js
const request = nativeRequest.call(transport, {
  ...options,
  method,
  headers: { ...headers },
  ...NodeURL.parse(pollyRequest.url)
});
```

packages/@pollyjs/adapter-node-http/src/utils/transport-wrapper.js
yasinuslu: I saw your implementation, i understand now. I was afraid to break the original request but as long as we bind user's event handlers we're ok 😄

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
offirgolan: Yeah it was 3am when I was working on that part lol. I'll fix it 😄

packages/@pollyjs/adapter-node-http/tests/node/integration/adapter-test.js
```diff
@@ -1,12 +1,73 @@
...
+import crypto from 'crypto';
+
 import setupFetchRecord from '@pollyjs-tests/helpers/setup-fetch-record';
 import adapterTests from '@pollyjs-tests/integration/adapter-tests';
 import FSPersister from '@pollyjs/persister-fs';
 import { setupMocha as setupPolly } from '@pollyjs/core';
 
 import NodeHttpAdapter from '../../../src';
 
+const collectBuffer = stream => {
```
yasinuslu: I wasn't sure where to put these util methods so i left them here

packages/@pollyjs/adapter-node-http/tests/node/integration/adapter-test.js
```diff
@@ -21,4 +82,16 @@ describe('Integration | Node Http Adapter | Node', function() {
...
 
   adapterTests();
+
+  binaryDownloadTest(
+    'should be able to fetch binary content with http',
+    http,
+    'http://via.placeholder.com/150/92c952'
+  );
+
+  binaryDownloadTest(
```
yasinuslu: I also want to add upload tests, i'm not sure where can i upload them. Should i create an express server in tests just to test uploading ?

yasinuslu:
CI is complaining about old commit messages, should i rewrite the history ?
image

packages/@pollyjs/adapter-node-http/tests/node/integration/adapter-test.js
offirgolan: Yeah, you can just modify `tests/middleware.js` and add your own upload route.

offirgolan:
> CI is complaining about old commit messages, should i rewrite the history ?

Yeah go for it.

timpur:
@yasinuslu i cant get this to work with https://github.com/request/request

it records the request and says it replays, but the request till fails when the local dev server is shutdown, wondering you you could give it a test ?

offirgolan:
@timpur I'm not able to recreate the issue you're having. Can you provide some more details on the issue you're facing?

packages/@pollyjs/adapter-node-http/src/-private/http-wrapper.js
```diff
@@ -0,0 +1,38 @@
...
+        new TransportWrapper(http, { name: 'http', adapter })
+      );
+    }
+
+    if (
+      options.transports.includes('https') &&
+      // Node 8 and below use http.request under the hood for https.request
+      // so we should skip wrapping it even if enabled.
+      // https://github.com/nodejs/node/blob/v8.14.0/lib/https.js#L245
+      semver.gte(process.version, '9.0.0')
```
offirgolan: @jasonmit @yasinuslu FYI. Im not sure if we should just throw here and force the user to to remove `https` from the transports list instead. I opted for a silent skip since this will be removed once we not longer support Node 6 & 8.

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -0,0 +1,290 @@
...
+      request: this.transport.request,
+      get: this.transport.get
+    });
+
+    this.transport.request = this.createRequestWrapper();
+
+    // In Node 10+, http.get no longer references http.request by the export
+    // so we need to make sure we wrap it as well.
+    // https://github.com/nodejs/node/blob/v10.0.0/lib/https.js#L275
+    this.transport.get = this.createGetWrapper();
```
offirgolan: @yasinuslu FYI

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -0,0 +1,290 @@
...
+        if (chunk) {
+          req.write(chunk, encoding);
+        }
+
+        // No need to carry callback around. This is what happens in original `end`.
+        if (typeof callback === 'function') {
+          req.once('finish', callback);
+        }
+
+        const headers =
```
offirgolan: @yasinuslu I've removed the `response` listener removal here since this request never actually gets sent out. This allows us to keep all listeners intact and emit the events ourselves in `respond`.

packages/@pollyjs/adapter-node-http/src/-private/http-wrapper.js
yasinuslu: I think you're right about silently skipping, maybe we can add it to the readme as a warning

docs/adapters/node-http.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Usage
+
+Use the [configure](api#configure), [connectTo](api#connectto), and
+[disconnectFrom](api#disconnectfrom) APIs to connect or disconnect from the
+adapter.
+
+```js
+import { Polly } from '@pollyjs/core';
+import NodeHttpAdapter from '@pollyjs/adapter-fetch';
```
jasonmit: I think this import is wrong

packages/@pollyjs/adapter-node-http/package.json
```diff
@@ -0,0 +1,60 @@
...
+  ],
+  "keywords": [
+    "polly",
+    "pollyjs",
+    "record",
+    "replay",
+    "http",
+    "adapter"
+  ],
+  "publishConfig": {
```
jasonmit: Should this be here?

packages/@pollyjs/adapter-node-http/src/-private/http-wrapper.js
jasonmit: Agree with @yasinuslu, seems fine just needs documented.

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -0,0 +1,290 @@
...
+
+    nativeRequestMapping.delete(this.transport);
+  }
+
+  getBodyFromChunks(chunks, headers) {
+    const { adapter } = this;
+
+    //  If we have headers and there is content-encoding it means that
+    //  the body shouldn't be merged but instead persisted as an array
+    //  of hex strings so that the responses can be mocked one by one.
```
jasonmit: Here is my take on an alternative (took me a few minutes to understand what this meant):

```js
//  If content-encoding is set in the header then the body/content
//  should not be concatenated.  Instead, the chunks should
//  be preserved as-is so that each chunk can be mocked individually
```

packages/@pollyjs/adapter-node-http/tests/utils/calculate-hash-from-stream.js
```diff
@@ -0,0 +1,12 @@
+import crypto from 'crypto';
+
+import getBufferFromStream from './get-buffer-from-stream';
+
+export default async function calculateHashFromStream(stream) {
+  const hmac = crypto.createHmac('sha256', 'a secret');
+  const hashStream = stream.pipe(hmac);
+
```
jasonmit: nit whitespace

packages/@pollyjs/adapter-node-http/package.json
offirgolan: Yeah, thats how lerna knows that the package is public and should be published.

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -0,0 +1,292 @@
...
+
+    const response = await new Promise(resolve => {
+      request.once('response', response => resolve(response));
+      request.end();
+    });
+
+    const responseBody = await new Promise(resolve => {
+      const chunks = [];
+
+      response.on('data', chunk => chunks.push(chunk));
```
jasonmit: Should we need to handle the `error` event?

packages/@pollyjs/adapter-node-http/package.json
jasonmit: TIL.  Sounds similar to pkg.json `"private"`

offirgolan: Yeah, I'm not sure why they decided to have their own config for this...

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
offirgolan: Done 😄

offirgolan:
@yasinuslu released in `v1.4.0` 🎉!

yasinuslu:
Thank you @offirgolan, sorry I wasn't able to participate more

offirgolan:
No worries at all @yasinuslu! Thank you for putting so much time into this. This wouldnt have gotten done so quickly if it wasnt for you 😄.

timpur:
@offirgolan sorry on holls ... will test out and report back, thanks guys :)

timpur:
@offirgolan did you ever test a POST request using https://github.com/request/request, i get `Error: connect ECONNREFUSED 127.0.0.1:4000` even though polly responds with `Replayed ➞ POST http://localhost:4000/`

jasonmit:
@timpur can you open a separate issue for visibility

timpur:
Sorry, https://github.com/Netflix/pollyjs/issues/156

PR #130, opened 2018-11-09
test: Fix hanging jest test case

offirgolan:


jest.config.js
```diff
@@ -3,5 +3,6 @@
 module.exports = {
   testURL: 'http://localhost:4000/api',
   testMatch: ['**/@pollyjs/*/build/jest/*.js'],
+  roots: ['<rootDir>/packages/@pollyjs'],
```
offirgolan: Tests were hanging because jest was registering the node-fetch mock in the examples directory. This scopes down the root to only look for files in the pollyjs packages directories.

PR #131, opened 2018-11-10
chore: Bump dev dependencies

offirgolan:
- Bump all the packages!
- Cleanup eslint config and move lint commands to top level


build-scripts/rollup.browser.config.js
```diff
@@ -38,6 +38,7 @@ export default function createBrowserConfig(options = {}, targets) {
           plugins: [
             '@babel/plugin-external-helpers',
             ['@babel/plugin-transform-runtime', { corejs: 2 }],
+            ['@babel/plugin-proposal-class-properties', { loose: true }],
```
offirgolan: This allows us to use class properties. I'll start migrating over to use them in another PR and include better inline documentation.

PR #134, opened 2018-11-11
chore: Remove class properties babel plugin and babel-eslint

offirgolan:
In order to support class properties, we would have to make the commonjs rollup plugin run after babel which isnt something I want to mess with atm. This PR reverts the setup needed to use class-properties both with rollup and eslint.


examples/dummy-app/.eslintrc.js
```diff
@@ -1,3 +1,8 @@
 module.exports = {
-  extends: ['plugin:react/recommended']
+  extends: ['plugin:react/recommended'],
+  settings: {
+    react: {
+      version: '16.5.1'
```
offirgolan: Needed to add a react version here to stop an eslint warning.

.eslintrc.js
```diff
@@ -1,11 +1,10 @@
 module.exports = {
   root: true,
-  parser: 'babel-eslint',
   parserOptions: {
-    ecmaVersion: 2017,
+    ecmaVersion: 2018,
```
offirgolan: Upping the ECMA version here allows for Object spread support.

packages/@pollyjs/ember/.eslintrc.js
```diff
@@ -4,18 +4,12 @@ module.exports = {
     ecmaVersion: 2017,
     sourceType: 'module'
   },
-  plugins: [
```
offirgolan: This entire file is just prettier auto-formatting.

PR #138, opened 2018-11-19
feat(core): Support custom functions in matchRequestsBy config options

### LINKED ISSUE #136: Ignore URL parameter for the comparison?
Hello,

I was wondering is there is a way to ignore a particular parameter in the request so that when doing the comparison to see if the lib store or record the call it is replayed even is the parameter is different.

I mean in our webapp we have a parameter that depends on the session and that will also be different if we redo the exact same call let's say. 
I'm trying to use the lib to ease frontend development by abstracting our backend. We will record scenario/ flows and will replay them without a real backend in place.

Let me know if it's possible to ignore a parameter regarding the comparison.

Maxime.

offirgolan:
Resolved #136.

Add more fine grain control on how request ids are generated by allowing custom methods to be passed into the `matchRequestsBy` config options.

### DIFF
--- a/packages/@pollyjs/core/src/-private/request.js
+++ b/packages/@pollyjs/core/src/-private/request.js
@@ -129,11 +129,8 @@ export default class PollyRequest extends HTTPBase {
     this.response = new PollyResponse();
     this.didRespond = false;
 
-    // Serialize the body which handles FormData + Blobs/Files
-    this.serializedBody = await this.serializeBody();
-
     // Setup this request's identifiers, id, and order
-    this._identify();
+    await this._identify();
 
     // Timestamp the request
     this.timestamp = timestamp();
@@ -172,10 +169,6 @@ export default class PollyRequest extends HTTPBase {
     await this._emit('response', this.response);
   }
 
-  async serializeBody() {
-    return serializeRequestBody(this.body);
-  }
-
   _overrideRecordingName(recordingName) {
     validateRecordingName(recordingName);
     this.recordingName = recordingName;
@@ -195,7 +188,7 @@ export default class PollyRequest extends HTTPBase {
     return this[ROUTE].emit(eventName, this, ...args);
   }
 
-  _identify() {
+  async _identify() {
     const polly = this[POLLY];
     const { _requests: requests } = polly;
     const { matchRequestsBy } = this.config;
@@ -205,11 +198,15 @@ export default class PollyRequest extends HTTPBase {
     keys(NormalizeRequest).forEach(key => {
       if (this[key] && matchRequestsBy[key]) {
         identifiers[key] = NormalizeRequest[key](
-          key === 'body' ? this.serializedBody : this[key],
+          this[key],
           matchRequestsBy[key]
         );
       }
     });
+
+    if (identifiers.body) {
+      identifiers.body = await serializeRequestBody(identifiers.body);
+    }
 
     // Store the identifiers for debugging and testing
     this.identifiers = freeze(identifiers);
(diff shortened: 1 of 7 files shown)


packages/@pollyjs/core/src/-private/request.js
```diff
@@ -205,12 +198,16 @@ export default class PollyRequest extends HTTPBase {
     keys(NormalizeRequest).forEach(key => {
       if (this[key] && matchRequestsBy[key]) {
         identifiers[key] = NormalizeRequest[key](
-          key === 'body' ? this.serializedBody : this[key],
+          this[key],
```
offirgolan: Pass the actual request body into the normalizer. We will serialize the the returned body after.

packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -8,15 +9,23 @@ const { keys } = Object;
 const { isArray } = Array;
 const { parse } = JSON;
 
-export function method(method) {
-  return (method || 'GET').toUpperCase();
+export function method(method, config) {
```
offirgolan: Not really sure if this is actually necessary....

packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -31,18 +40,23 @@ export function headers(headers, config) {
...
+    config.exclude.forEach(header => delete normalizedHeaders[header]);
   }
 
   return normalizedHeaders;
 }
 
-export function body(body) {
-  return body;
+export function body(body, config) {
+  return isFunction(config) ? config(body) : body;
```
offirgolan: The body passed here is the actual request body. In most cases it will be a string but it could also be a FormData, Blob, or maybe even a Buffer instance so we can't really add any preventative logic here to clone the body. 

The only solution I can think of is adding a warning in the docs about this.

packages/@pollyjs/core/src/utils/normalize-request.js
```diff
@@ -1,4 +1,5 @@
 import isObjectLike from 'lodash-es/isObjectLike';
+import isFunction from 'lodash-es/isFunction';
```
jasonmit: nit: Don't think we need lodash for this

packages/@pollyjs/core/src/utils/normalize-request.js
jasonmit: > The body passed here is the actual request body. In most cases it will be a string but it could also be a FormData, Blob, or maybe even a Buffer instance so we can't really add any preventative logic here to clone the body.

The fear here is someone mutating the body where it would effect more than just the matching comparison?

packages/@pollyjs/core/src/utils/normalize-request.js
jasonmit: is method not always provided which is why we have the fallback?  Should we fix that upstream?

packages/@pollyjs/core/src/utils/normalize-request.js
offirgolan: Correct. We pass the actual request body that was originally passed in and its done before the request is sent out so if someone were to modify the Blob or FormData instance then that would be the one that would get sent out with the request. 

We can just turn this off if we think its too sketchy (along with `method`) and just support url type methods.

packages/@pollyjs/core/src/utils/normalize-request.js
offirgolan: It is 😄. Updated.

packages/@pollyjs/core/src/utils/normalize-request.js
jasonmit: > We can just turn this off if we think its too sketchy.

I see some value in being able to swap parts of the body used in the comparison (graphql requests).

offirgolan: Ah right, makes perfect sense. I can just add a warning in the docs. Was there something specific you wanted to add/modify?

jasonmit: Brought the conversation to slack

PR #140, opened 2018-11-25
feat: Add an onIdentifyRequest hook to allow adapter level serialization

offirgolan:
This PR adds an `onIdentifyRequest` hook for all adapters which will allow them to each handle the serialization of certain identifiers (mostly the request body). I've created a `blob`, `formData`, and `buffer` serializer that can be used or an adapter may choose to implement its own. By default, all 3 will be run as they are each safe to be executed in both environments.


packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -81,6 +83,16 @@ export default class FetchAdapter extends Adapter {
...
   }
 
+  async onIdentifyRequest(pollyRequest) {
+    const { identifiers } = pollyRequest;
+    let body = identifiers.body;
+
+    body = await BrowserSerializers.serializeRequestBody(body);
+    body = NodeSerializers.serializeRequestBody(body);
+
+    identifiers.body = body;
```
offirgolan: @jasonmit since both serializers are browser and node safe, should we just default to this logic on the core Adapter class and adapters that want to override this will have the ability to?

jasonmit: >  override this

That seems fine.  Unrelated, I'm confused by the browser and node serializer being invoked here.

offirgolan: After discussions on slack, we've decided to split the serializers by type (not environment) and run them all by default.

PR #160, opened 2019-01-11
fix(adapter-node-http): Fix unhandled rejection if connection fails

dustinsoftware:
An unhandled promise rejection was thrown if a socket could not be opened,
which prevents handling the error gracefully in a consuming app.

First of all, thanks for building this library! I'm exposing a HTTP API on top of it [here](https://github.com/dustinsoftware/polly-proxy/) and it's working pretty well. However..

I'm running into a crash when polly can't connect to an API (port isn't open, for instance).

In my app, I'm tearing down the process if there are any unhandled promise rejections by using:

```
process.on('unhandledRejection', exception => {
	throw exception;
});
```

I had to fix promise rejection handling in a few places, but it works properly now. I've confirmed that this fix allows polly-proxy to return a graceful HTTP 500 from the express instance instead of crashing the app, but am unsure the best way to add a test for this type of bug.


dustinsoftware:
Looks like CI passed for Node 8 and 10, but failed for 6 during package restore 🤷‍♂️

dustinsoftware:
I've added a test that verifies the expected behavior in [polly-proxy](https://github.com/dustinsoftware/polly-proxy/tree/test-updates) on the `test-updates` branch.

To repro with that project:
`yarn build && yarn test:ci`

To verify this fix:
`yarn link @pollyjs/adapter-node-http`
`yarn link @pollyjs/core`
`yarn build && yarn test:ci`

I'm not sure that adding `promise.catch(() => {})` is the best solution here, although it doesn't seem to negatively affect consumers (such as `flush`)

dustinsoftware:
I've added a test that verifies the expected behavior in [polly-proxy](https://github.com/dustinsoftware/polly-proxy/tree/test-updates) on the `test-updates` branch.

To repro with that project:
`yarn build && yarn test:ci`

To verify this fix:
`yarn link @pollyjs/adapter-node-http`
`yarn link @pollyjs/core`
`yarn build && yarn test:ci`

I'm not sure that adding `promise.catch(() => {})` is the best solution here, although it doesn't seem to negatively affect consumers (such as `flush`)

dustinsoftware:
FWIW I was able to successfully hack around the issue in `deferred-promise.js` but ended up forking `adapter-node-http` until we can get a fix in :) 

https://github.com/dustinsoftware/polly-proxy/commit/c558c6cc272d0129199b62ebf54a0a700606f152

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -138,15 +138,12 @@ export default class TransportWrapper {
       ...NodeUrl.parse(pollyRequest.url)
     });
 
-    request.on('error', error => {
```
offirgolan: Whats the reason for removing this? By placing this here, we can throw on errors emitted during writing chunks to the request.

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -270,13 +267,17 @@ export default class TransportWrapper {
...
+        adapter
+          .handleRequest({
+            method,
+            headers,
+            url: parsedUrl.href,
+            body: chunks,
+            requestArguments: [wrapper, req, args]
+          })
+          .catch(e => {
+            req.emit('error', e);
```
offirgolan: Can you add you commit message as an inline message so future readers will understand why this is necessary? 

```
This allows the consumer to handle the error gracefully (there was no way to detect this error previously)
```

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
dustinsoftware: The error subscription was moved to line 146 so that it could reject the promise instead of throwing an unhandled error.  I'll try adding the subscription before calling `request.write` so we don't miss out on errors writing the request body.

offirgolan: Im not entirely sure what the difference in outcome is with the change. Both will result in the async method's promise to be rejected which you've caught on line 280.

dustinsoftware: The key difference appears to be throwing an error inside the callback function. This causes the `uncaughtException` event to fire because the thrown error is not attributed to the current async state in `passthrough`, as the error callback is invoked asynchronously when the 'error' event is fired.

I've made a simple express instance that repro's the behavior before and after the proposed change here: 
https://glitch.com/edit/#!/voracious-pilot

offirgolan: AHHH! Ok I see it now! Thanks for the clarification.

packages/@pollyjs/adapter-node-http/src/-private/transport-wrapper.js
```diff
@@ -138,17 +138,16 @@ export default class TransportWrapper {
...
     });
 
     // Write the request body
     chunks.forEach(chunk => request.write(chunk));
 
-    const response = await new Promise(resolve => {
-      request.once('response', response => resolve(response));
-      request.end();
-    });
+    request.end();
```
offirgolan: What was the need to move this out of the above promise?

dustinsoftware: The request is ended prematurely, because the call to end happens before the request body has been written. Luckily this caused a test to fail :)

PR #166, opened 2019-01-25
 feat(adapter-node-http): Use `nock` under the hood instead of custom implementation

offirgolan:
The reasoning behind this drastic change in implementation is that although we were able to override request methods  once it was created, we did not take into account the socket creation that was setup in the initialization of the ClientRequest. To be able to modify that logic, we would have to mock out the entire ClientRequest class which nock already does and does it well.


packages/@pollyjs/adapter-node-http/src/index.js
```diff
@@ -1,49 +1,273 @@
...
+    const modules = { http, https };
+    const { ClientRequest } = http;
+
+    // Patch the already overridden ClientRequest class so we can get
+    // access to the original arguments and use them when creating the
+    // passthrough request.
+    http.ClientRequest = function _ClientRequest() {
+      const req = new ClientRequest(...arguments);
+
+      req[ARGUMENTS] = [...arguments];
```
jasonmit: Would a WeakMap work here?

tests/recordings/-pollyjs_2160168770/adapter-node-http_1501306138/Integration-Node-Http-Adapter_1543864754/http_3378792613/should-call-all-the-life-cycle-events_2199659288/recording.har
```diff
@@ -23,9 +23,6 @@
           "headersSize": 230,
           "httpVersion": "HTTP/1.1",
           "method": "GET",
-          "postData": {
```
jasonmit: If it's plain we no longer need to store the mimeType?

packages/@pollyjs/adapter-node-http/src/index.js
```diff
@@ -1,49 +1,273 @@
...
+      ...options,
+      method,
+      headers: { ...headers },
+      ...URL.parse(pollyRequest.url)
+    });
+
+    const chunks = this.getChunksFromBody(body, headers);
+
+    const responsePromise = new Promise((resolve, reject) => {
+      request.once('response', response => resolve(response));
```
jasonmit: Should we handle aborts and timeouts?

dustinsoftware:
@offirgolan would you mind publishing the latest changes on master? I'd like to beta test this unmerged PR but am having dependency issues since the adapter logic changed in `@pollyjs/adapter`.. (or could you publish this PR as an alpha?)

PR #167, opened 2019-01-25
chore: Remove support for Node 6

offirgolan:
Node 6 is coming to EOL on April 2019 and our build is already failing due to dependencies dropping support for it as well. To allow simpler implementations on the Node side of Polly, I believe we should drop support of Node 6 as well as it will no longer be maintained in ~3 months.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - '6'
   - '8'
   - '10'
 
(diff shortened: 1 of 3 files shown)


Turbo87:
@offirgolan you might want to update https://github.com/Netflix/pollyjs/blob/master/packages/%40pollyjs/ember/package.json#L78 too 😉

offirgolan:
Good call @Turbo87, ty. Fixed in #169.

PR #170, opened 2019-01-29
chore: Add script to check for uncommitted files after tests run

offirgolan:
This is to make sure that no files are modified, created, or removed (e.g. recordings) that havent already been checked into git.

_Original script taken from [here](https://stackoverflow.com/a/3879077)._


offirgolan:
The failed build is the reason for this PR. Here an the output:

screen shot 2019-01-28 at 8 43 55 pm

I noticed this when I was running the test suite locally

PR #172, opened 2019-01-29
fix(adapter): Log information if request couldn't be found in recording

### LINKED ISSUE #114: Not enough information when recording isn't found
Pollyjs only tells the URL and method when a recording isn't found for a request. It's especially difficult to tell which request wasn't matched when using GraphQL, because all requests have the same method and URL. An ideal enhancement would be if it told you why it doesn't match, but just an option for more verbose logging of the missing recording would be very helpful!

dustinsoftware:
These three fields are referenced in the _identify function.

Fixes #114

### DIFF
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -111,8 +111,16 @@ export default class Adapter {
   async [REQUEST_HANDLER](pollyRequest) {
     const { mode } = this.polly;
     let interceptor;
+    const {
+      shouldIntercept,
+      shouldPassthrough,
+      identifiers,
+      id,
+      order,
+      url
+    } = pollyRequest;
 
-    if (pollyRequest.shouldIntercept) {
+    if (shouldIntercept) {
       interceptor = new Interceptor();
       const response = await this.intercept(pollyRequest, interceptor);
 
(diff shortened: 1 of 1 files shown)


dustinsoftware:
1 test appears to be a flake (adapter-fetch test timed out)

offirgolan:
@dustinsoftware there are a couple of other places where we log

```js
`${pollyRequest.method} ${pollyRequest.url}`
```

Can you replace all of those as well? A small utility method to stringify a request for logging would be good here 😄

dustinsoftware:
Sure thing.

dustinsoftware:
Hmm. Just did a quick search, this is the only other use I can find. Won't this exception only be thrown if there is a bug in polly itself?

```js
    // This should never be reached. If it did, then something screwy happened.
    this.assert(
      `Unhandled request: ${pollyRequest.method} ${pollyRequest.url}.`,
      false
    );
```

offirgolan:
Oh... I totally thought there were a few other places but I was mistaken. My bad! Lets just add it there as well for consistency. If it does ever reach that point, at least the user will have more information about the request.

dustinsoftware:
Updated, let me know if there's any other things you'd like changed :)

packages/@pollyjs/adapter/src/index.js
```diff
@@ -205,7 +213,8 @@ export default class Adapter {
 
     this.assert(
       'Recording for the following request is not found and `recordIfMissing` is `false`.\n' +
-        `${pollyRequest.method} ${pollyRequest.url}\n`,
+        JSON.stringify({ identifiers, id, order }, null, 2) +
```
offirgolan: We should add `url` as well since identifiers might not have all the details needed depending on the user's config (e.g. matchRequestsBy).

PR #179, opened 2019-02-02
fix: Correctly handle array header values

offirgolan:
Correctly serialize and deserialize array header values to and from the stored HAR file.


packages/@pollyjs/adapter/src/utils/normalize-recorded-response.js
```diff
@@ -0,0 +1,30 @@
...
+    headers: normalizeHeaders(headers),
+    body: content && content.text
+  };
+}
+
+function normalizeHeaders(headers) {
+  return (headers || []).reduce((accum, { name, value }) => {
+    const existingValue = accum[name];
+
+    if (existingValue) {
```
offirgolan: The one gotcha about this is that if the original header value was `['foo']`, it would create a single header name/value pair so when we parse the recorded entry again, the value will just be `'foo'`.

That means the expected value type would change between the first and second run. We could move this logic into the Headers class so it would do the following: 

```js
headers.foo = ['bar']
headers.foo // -> 'bar'

headers.foo = ['bar', 'baz']
headers.foo // -> ['bar', 'baz']
```

But that would cause another gotcha:

```js
headers.foo = ['bar']
headers.foo.push('baz') --> throws cause headers.foo is a string
```

PR #182, opened 2019-02-13
refactor(adapter-puppeteer): Use respondToRequest and do not no-op hooks

offirgolan:


packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -148,7 +140,14 @@ export default class PuppeteerAdapter extends Adapter {
   }
 
   onRequest(pollyRequest) {
-    const [request] = pollyRequest.requestArguments;
+    const { request } = pollyRequest.requestArguments;
+    const { promise } = pollyRequest;
+
+    // Override the deferred promise's resolve and reject to no-op since
+    // we handle it manually in the `requestfinished` and `requestfailed` events.
+    promise._resolve = promise.resolve;
```
offirgolan: The `onRequestFinished` and `onRequestFailed` hooks used to only contain logic to resolve and reject the deferred promise. This has now changed and I continue to need to add some more logic to those hooks. Instead of no-oping the hooks, we can instead do it to the promise and keep it much more contained.

packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -220,6 +210,17 @@ export default class PuppeteerAdapter extends Adapter {
     }
   }
 
+  respondToRequest(pollyRequest) {
```
offirgolan: This public hook was introduced in the base adapter refactoring so just moving the related logic here.

PR #185, opened 2019-02-19
feat: Add `error` event and improve error handling

### LINKED ISSUE #181: Need info to be logged when record not found
Hello guys,
Pollyjs is throwing an error when a record is not found. In my use case it's not enough.

I am using pollyjs to record and replay requests in an application. In the application I have a general mechanism to catch all the errors and this is capturing also the errors thrown when the record is not found which is the expected behavior.
The pollyjs instance is created on a separate app used only to monitor the requests. So the problem is here, because I want to log an info message when a record is not found. For the request which are found I can listen to the 'beforeReplay' event so I can log the information. 
Would it be possible to emit also a message when a record is not found in the 'beforeReplay' event (For the moment 'beforeReplay' emits only when a recordingEntry is found)? 
Or, is there other possible solution to be able to get an information when the record is not found (other than the error which is thrown) ?

offirgolan:
Resolves #181 

```js
server.any().on('error', (req, error) => {
  console.error(error);
  process.exit(1);
});
```

### DIFF
--- a/packages/@pollyjs/adapter-puppeteer/src/index.js
+++ b/packages/@pollyjs/adapter-puppeteer/src/index.js
@@ -202,6 +202,9 @@ export default class PuppeteerAdapter extends Adapter {
     const { response } = pollyRequest;
 
     if (error) {
+      // If an error was returned then we force puppeteer to abort the current
+      // request. This will emit the `requestfailed` page event and allow the end
+      // user to handle the error.
       await request.abort();
     } else {
       await request.respond({
--- a/packages/@pollyjs/adapter-puppeteer/src/index.js
+++ b/packages/@pollyjs/adapter-puppeteer/src/index.js
@@ -156,18 +156,6 @@ export default class PuppeteerAdapter extends Adapter {
     this._requestsMapping.pollyRequests.set(request, pollyRequest);
   }
 
-  /**
-   * Abort the request on failure.
-   *
-   * @override
-   */
-  async onRequestFailed(pollyRequest) {
-    const { request } = pollyRequest.requestArguments;
-
-    await request.abort();
-    await super.onRequestFailed(...arguments);
-  }
-
   async passthroughRequest(pollyRequest) {
     const { page } = this.options;
     const { id, order, url, method, headers, body } = pollyRequest;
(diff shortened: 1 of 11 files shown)


packages/@pollyjs/adapter/src/index.js
```diff
@@ -300,7 +300,12 @@ export default class Adapter {
    * @param {PollyRequest} pollyRequest
    * @param {Error} error
    */
-  onRequestFailed(pollyRequest, error) {
+  async onRequestFailed(pollyRequest, error) {
+    error =
+      error || new Error('[Polly] Request failed due to an unknown error.');
+
+    await pollyRequest._emit('error', error);
+    await this.respondToRequest(pollyRequest, error);
```
offirgolan: Call `respondToRequest` with an error if the request failed so that they adapters can gracefully handle it.

packages/@pollyjs/core/src/-private/logger.js
```diff
@@ -13,10 +13,6 @@ export default class Logger {
     this.groupName = null;
   }
 
-  get enabled() {
```
offirgolan: Removing dead/unused code.

packages/@pollyjs/adapter-xhr/tests/utils/xhr-request.js
```diff
@@ -22,7 +22,7 @@ export default function request(url, obj = {}) {
       xhr.status === 204 && xhr.responseText === '' ? null : xhr.responseText;
 
     return new Response(responseBody, {
-      status: xhr.status,
+      status: xhr.status || 500,
```
jasonmit: I realize this is just pertaining to a test util but can the fallback mask a future bug in our test suite?

packages/@pollyjs/adapter/src/utils/stringify-request.js
```diff
@@ -1,4 +1,12 @@
 export default function stringifyRequest(req, ...args) {
+  const config = { ...req.config };
+
+  // Remove all adapter & persister config options as they can cause a circular
+  // structure to the final JSON
+  ['adapter', 'adapterOptions', 'persister', 'persisterOptions'].forEach(
```
jasonmit: Is this a fix for separate bug?

packages/@pollyjs/adapter-puppeteer/src/index.js
```diff
@@ -203,22 +191,25 @@ export default class PuppeteerAdapter extends Adapter {
...
     const { request } = pollyRequest.requestArguments;
     const { response } = pollyRequest;
 
-    request.respond({
-      status: response.statusCode,
-      headers: response.headers,
-      body: response.body
-    });
+    if (error) {
+      await request.abort();
```
jasonmit: Curious why the abort for the puppeteer adapter?

jasonmit: Might be worth leaving a comment on why we abort here.  It's not clear and realize it's code that was just moved in this PR.

packages/@pollyjs/adapter-xhr/tests/utils/xhr-request.js
offirgolan: This will only happen when status is `0` which occurs when we do `xhr.error()`. Do you have a better way around this? Response wont accept a status code of `0`

packages/@pollyjs/adapter/src/utils/stringify-request.js
offirgolan: This is a fix for a separate bug that I found when tackling this.

PR #202, opened 2019-04-18
feat(core): Improved control flow with `times` and `stopPropagation`

offirgolan:
## Description

This PR includes 2 features that when used together, can be really powerful. 

The first is the ability to remove an intercept or event handler after a specified amount of times it has been called via the handler's `.times(n)` API or by providing it as an option to `.on(eventName, fn, { times })` or `.intercept(fn, { times })`.

The second, is the ability to prevent calling the next handlers registered for the events or intercept. This can be done via `interceptor.stopPropagation()` or via a new Event instance passed to every event handler by calling `event.stopPropagation()`.

## Motivation and Context

Lets take an example where we might want to intercept a user being removed. The first call to GET the user should return the user's info but after we delete the user, it should return a 404.

```js
// First call should return the user and not enter the 2nd handler
server
  .get('/users/1')
  .times(1) // Remove this interceptor after it gets called once
  .intercept((req, res, interceptor) => {
    // Do not continue to the next intercept handler which handles the 404 case
    interceptor.stopPropagation(); 
    res.sendStatus(200);
  });

server.delete('/users/1').intercept((req, res) => res.sendStatus(204));

// Second call should 404 since the user no longer exists
server
  .get('/users/1')
  .intercept((req, res) => res.sendStatus(404));
```

Then when the following requests get made, we are able to simulate the correct responses:

```js
await fetch('/users/1'); // --> 200
await fetch('/users/1', { method: 'DELETE' }); // --> 204
await fetch('/users/1');  // --> 404
```

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/adapter-fetch/tests/integration/server-test.js
+++ b/packages/@pollyjs/adapter-fetch/tests/integration/server-test.js
@@ -201,6 +201,42 @@ describe('Integration | Server', function() {
         /Invalid filter callback provided/
       );
     });
+
+    it('.times()', async function() {
+      const { server } = this.polly;
+      let callCount = 0;
+
+      server
+        .get('/ping')
+        .times(1)
+        .on('request', () => callCount++)
+        .times()
+        .intercept((req, res) => res.sendStatus(200));
+
+      expect((await fetch('/ping')).status).to.equal(200);
+      expect(callCount).to.equal(1);
+
+      expect((await fetch('/ping')).status).to.equal(200);
+      expect(callCount).to.equal(1);
+    });
+
+    it('.intercept(_, { times }) & .on(_, { times })', async function() {
+      const { server } = this.polly;
+      let callCount = 0;
+
+      const handler = server
+        .get('/ping')
+        .on('request', () => callCount++, { times: 1 })
+        .intercept((req, res) => res.sendStatus(200), { times: 2 });
+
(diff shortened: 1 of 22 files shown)


packages/@pollyjs/adapter-fetch/tests/integration/server-test.js
```diff
@@ -201,6 +201,42 @@ describe('Integration | Server', function() {
         /Invalid filter callback provided/
       );
     });
+
+    it('.times()', async function() {
```
jasonmit: Might be worth adding a test for the precedence between `times()` + `{ times }` when both are used

packages/@pollyjs/core/src/server/handler.js
```diff
@@ -11,7 +13,9 @@ export default class Handler extends Map {
     super();
 
     this.set('config', {});
+    this.set('defaultOptions', {});
```
jasonmit: Curious why the name defaultOptions?

offirgolan: Its more future proofing than anything else. Will allow us to add more API to modify default options passed to the interceptor/events.

packages/@pollyjs/adapter-fetch/tests/integration/server-test.js
offirgolan: Done.

PR #212, opened 2019-05-29
chore(core): Remove unnecessary async fn

jasonmit:
Removing since it mangles the output in a very unexpected way - note the `_context.stop()` ends up within `beforeEach`.

image

### DIFF
--- a/packages/@pollyjs/core/tests/unit/test-helpers/mocha-test.js
+++ b/packages/@pollyjs/core/tests/unit/test-helpers/mocha-test.js
@@ -4,7 +4,7 @@ class Sandbox {
   constructor(context) {
     this.beforeEachCalled = new Set();
     this.afterEachCalled = new Set();
-    this.context = context || { currentTest: {} };
+    this.context = context || { currentTest: { title: 'mockname' } };
   }
 
   beforeEach(fn) {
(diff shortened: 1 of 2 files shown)


packages/@pollyjs/core/tests/unit/test-helpers/mocha-test.js
```diff
@@ -34,21 +34,15 @@ describe('Unit | Test Helpers | mocha', function() {
   });
 
   it('should create a polly property and set recordingName', function() {
```
jasonmit: The test on line 28 was the one that was now failing.  This was because the async was putting the [recording name assertion](packages/@pollyjs/core/src/utils/validators.js) on the next tick.

packages/@pollyjs/core/tests/unit/test-helpers/mocha-test.js
```diff
@@ -4,7 +4,7 @@ class Sandbox {
   constructor(context) {
     this.beforeEachCalled = new Set();
     this.afterEachCalled = new Set();
-    this.context = context || { currentTest: {} };
+    this.context = context || { currentTest: { title: 'mockname' } };
```
jasonmit: The test on line 28 (not visible) was the one that was now failing.  This was because the async was putting the [recording name assertion](https://github.com/Netflix/pollyjs/blob/master/packages/@pollyjs/core/src/utils/validators.js#L4) on the next tick.  I just needed to update the context object to give it a mock test name.

The other changes are me doing small clean ups on the test file for consistency.

offirgolan:
@jasonmit what is `_context` in the screenshot?

jasonmit:
> @jasonmit what is _context in the screenshot?

I don't recall exactly but I believe it was the "cleaned" up context object.

PR #218, opened 2019-06-10
chore: Bump dependencies

offirgolan:


jasonmit:
On my phone reviewing, do you know if any of these major dep bumps warrants a major on our end?  I.e., build targets changing

offirgolan:
Most of the major version upgrades remove support for Node 6. Im not sure if that warrants a major version upgrade for us or not as we've already dropped support for it but we can do a major version bump to be safe.

Issue #183, opened 2019-02-18
Support binary response data in XHR & Fetch adapters


offirgolan:
Properly support recording and responding with received binary data in the XHR and Fetch adapters.

badsyntax:
I'm not sure this helps at all, maybe it's helpful for others, but i'm able to work with binary requests with `@pollyjs/core@^1.4.1` and `@pollyjs/adapter-fetch@1.2.0`.

offirgolan:
@badsyntax binary requests or responses?

badsyntax:
@offirgolan The following works with the versions i've mentioned (ignoring the `cy.fixture` part, this is part of Cypress, and it's just a helper to get contents of a fixture file):

```javascript
cy.fixture('writeApiResponse.bin').then(fixture => {
  pollyServer.post('http://127.0.0.1:3000/binary').intercept((_, res) => {
    res.setHeader('Content-Type', 'application/x-protobuf');
    res.send(fixture);
  });
});
```

With `@pollyjs/core@^2.3.0` and `@pollyjs/adapter-fetch@^2.3.1`, an error is thrown on the following line:

```javascript
pollyServer.post('http://127.0.0.1:3000/binary').intercept((_, res) => {
```

The error is: 

```javascript
Uncaught (in promise) TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.
```

offirgolan:
@badsyntax this issue is in regards to adding proper support for binary responses from a server, not the intercept method (e.g. recording an image from the server).

Would you mind creating a separate issue with your previous comment so we can properly track it? If you can also include an example of what `fixture` is, that would be super useful.

badsyntax:
I got the impression it was related. I'll create a seperate ticket.

badsyntax:
Hi @offirgolan After spending some time getting a proper test case together to demonstrate this, i've realised that the error I mentioned is a result of using Polly.js within Cypress. The error originates form Polly code but could be due to something Cypress is doing. I'm unable to generate the error outside of Cypress. Really sorry for highjacking this issue. Would you still find it useful for me to get a test repo together (that uses Cypress) to replicate the error?

offirgolan:
> Would you still find it useful for me to get a test repo together (that uses Cypress) to replicate the error?

Yeah, I'm starting to work a bit more with cypress and would love a reproduction to play around in. If you can create a new issue with it that would be great!

[...]

PR #220, opened 2019-06-29
Mapbox GL fixes: handle `Request` objects as URLs (and array buffer responses discussion)

tombh:
## Description
These are fixes I needed to get Polly working with Mapbox GL. So far the handling of an array buffer response is only for passthrough requests, not recording, is that a reasonable first step? There are no tests because I'd like some help thinking about how to test these.

## Motivation and Context
I wonder if the handling of `Request` objects for URLs has already been considered? It's just that the the coercion of those objects into a URL string isn't handled simply by casting to a string?

I see in #183 that array buffer responses have already been considered.

## Types of Changes

## Checklist


tombh:
I think those test failures are linting errors on files not changed in this PR?

offirgolan:
@tombh thanks for taking the time in opening up this PR! 

> I wonder if the handling of Request objects for URLs has already been considered? It's just that the the coercion of those objects into a URL string isn't handled simply by casting to a string?

This is definitely something that we should handle and should've been considered but I guess was missed. Can you split this into a separate PR so we can get this in faster as the ArrayBuffer work is going to be a bit more time consuming. 

One note: `PollyRequest` does expect `url` to be a string or an object that can be casted into a url string. The logic to pass the correct url from a Request instance should be handled in the fetch adapter layer instead. 

> So far the handling of an array buffer response is only for passthrough requests, not recording, is that a reasonable first step?

Due to the current nature of the client server API (`polly.server`) even though we only target supporting binary data in passthroughs, it still flows through all the server route handlers and middleware. This can cause a pretty trolling behavior since the response body will be different depending on the recording mode. 

Im not exactly sure how you want to proceed with this but supporting binary data would need to work across all recording modes in order to be merged in.

tombh:
Ok I've removed the ArrayBuffer fix and added a test for the Request object fix. BTW is there not a way to run a single test? My laptop quite struggles with `testem`.

tombh:
Again tests seem to fail because of an unassociated linting errors. But in fact I don't even seem to be able to see my added test in the logs.

offirgolan:
Thanks for taking the time to land this @tombh. The CI failures are unrelated, we'll fix them separately. 

> BTW is there not a way to run a single test? My laptop quite struggles with testem.

Yeah, the current test/build system needs some work for sure. You can run a single test by using `it.only` on the test and then running `yarn test -l Chrome` or `yarn test:ci -l Chrome`.

packages/@pollyjs/core/src/-private/request.js
```diff
@@ -82,6 +82,9 @@ export default class PollyRequest extends HTTPBase {
   }
 
   set url(value) {
+    if (typeof value === 'object' && 'url' in value) {
```
offirgolan: This logic should be moved into the fetch adapter.

Issue #219, opened 2019-06-29
Clarification on use of Local Storage


tombh:
In the docs section [Using Polly In The Browser](https://netflix.github.io/pollyjs/#/quick-start?id=using-polly-in-the-browser) it suggests that Local Storage is an alternative option to using the Rest persister. The implication seems to be that a browser's local storage can replace all the functionality of the Rest persister, however my immediate reaction to that is that HARs stored in the browser cannot be committed to my repo and subsequently replayed in CI. Am I missing something here? I assume the Local Storage persister is more for working offline, not for building CI tests?

In starting to read that paragraph I knew I'd found the right option for my use case, but by the end I was left in confusion. Could there be a simple amendment to help clarify?

jasonmit:
Correct.  The local storage is primarily for local, offline, support.

It's definitely not intended to replace the use cases of the rest or fs persisters.  If there's language throwing you off in the docs, please send a PR our way to correct it 🙏

PR #223, opened 2019-07-04
chore(core): Improve "...in the Browser?" docs

tombh:
## Description
Rewriting of the "Using Polly in the Browser?" section.

## Motivation and Context
Just to improve clarity. See #219 for further details.

## Types of Changes
Just documentation

## Checklist

BTW, I've been surprised a couple of times by the pre-commit hook not saving my commit message on hook failure. I don't know what the solution is, but it'd be good to either know that the commit message won't necessarily be saved (and so need to be written again), or that something like a pre-push hook be used so the commit message is always available.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -68,23 +68,28 @@ new Polly('<Recording Name>', {
 
 ## Using Polly in the Browser?
 
-In order to write to disk from the browser, Polly will make networks requests to a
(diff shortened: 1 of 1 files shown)


jasonmit:
The CI failures are unrelated, I'll fix them separately.

jasonmit:
Thanks for taking the time to land this!

tombh:
Great, thanks :)

PR #225, opened 2019-07-05
ci: Fixes linting error for demo prop types

jasonmit:


examples/dummy-app/src/users.js
```diff
@@ -9,11 +10,16 @@ import {
   ShowButton
 } from 'react-admin';
 
-// eslint-disable-next-line react/prop-types
```
jasonmit: eslint-plugin-react wants this now on every use of the prop (so line 12 and line 14).  Could be a bug on their end but was simpler to just add the propTypes since it's already a dependency.

Issue #226, opened 2019-07-10
feature: Allow Polly throwing an error if recording is expired


sankaritan:
## Feature proposal: Allow configuring Polly to fail if recordings are expired.

We have successfully migrated a few of our projects to Polly - and it worked great so far! One feature we're currently missing is the ability to fail the test in case the recordings get out of date.

The current logic allows to either keep using expired mocks or re-record them when expired.
While this is usually satisfactory, it doesn't suit well our situation. Imagine:

1. The engineer starts working on a new feature and pushes changes to git without running tests. Ideally, this should never happen (pre-commit hook won't even let you!), but reality can differ, project from the project.
1. If recordings are out of date, they may be re-recorded on CI, but those updates would not be preserved in the source code. This can lead to a situation when recordings (contracts) slowly get out of date.

In order to be able to automatically enforce updating recordings as a hard rule before merging, I'd like to propose the following:

**A)** Introduce a new config option: `failIfExpired`, set to `false` by default. Call `this.asset("error text")` whenever recording is expired & the value of `failIfExpired` is set to `true`.

**B)** An alternative could be smth like `expiryMode` or `expiryStrategy` that would allow user to choose from multiple options:
- replay
- rerecord
- failure

Happy to make a PR if this sounds like something that would fit into Pollyjs roadmap!

offirgolan:
Thanks for opening up this issue @sankaritan! This enhancement makes a lot of sense and should be super easy to add. A PR would be greatly appreciated 😄.

For the sake of simplicity and backwards compatibility, I say we go with option **A**.

@jasonmit thoughts?

jasonmit:
I'm in favor but I'd prefer an event this way folks can choose how to implement it (throw, log, instrument, etc.)

jasonmit:
@sankaritan chatted with Offir we've aligned around `expiryStrategy`.

It will simplify the config and encapsulates two config options into one and eliminates confusion (i.e., `failIfExpired` is true and `recordIfMissing` is true).

I also believe we can achieve this in a backwards compatible way if we deprecate `recordIfMissing` and our config logic is updated to use the new strategy prop.

offirgolan:
Resolved in #227.

PR #227, opened 2019-07-12
 feat: Implement expiryStrategy

fastfrwrd:
## Description

Per discussion in #226, this implements an `expiryStrategy` config option and deprecates `recordIfExpired` (without a breaking change) . It supports the existing flows and adds the new `error` state, which throws when the strategy is set to `error` and an expired recording is about to be used in `replay` mode.

We are still implementing tests, but would love input on the approach.

## Motivation and Context

See #226 for detailed info.

## Types of Changes

## Checklist

- [x] I have added tests to cover my changes. - *Tests in progress, the `it` blocks are included for now, let us know if they would be in line with the overall testing strategy for the repo.*

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -54,8 +54,9 @@ polly.configure({
 });
 ```
 
-## recordIfExpired - _deprecated! use expiryStrategy_
+## recordIfExpired
 
+_deprecated! use [expiryStrategy](#expiryStrategy)_
 _Type_: `Boolean`
 _Default_: `false`
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -54,7 +54,7 @@ polly.configure({
 });
 ```
 
-## recordIfExpired
+## recordIfExpired - _deprecated! use expiryStrategy_
 
 _Type_: `Boolean`
 _Default_: `false`
@@ -93,7 +93,8 @@ _Type_: `String`
 _Default_: `null`
 
 After how long the recorded request will be considered expired from the time
-it was persisted.
+it was persisted. A recorded request is considered expired if the recording's
+`startedDateTime` plus the current `expiresIn` duration is in the past.
 
 **Example**
 
@@ -104,6 +105,25 @@ polly.configure({
 
 polly.configure({
   expiresIn: '5 min 10 seconds 100 milliseconds' // expires in 5 minutes, 10 seconds, and 100 milliseconds
+});
+```
+
+## expiryStrategy
+
+_Type_: `String`
+_Default_: `warn`
+
+The strategy for what should occur when Polly tries to use an expired recording in `replay` mode. Can be one of the following:
+
+- `warn`: Log a console warning about the expired recording.
+- `error`: Throw an error.
+- `record`: Re-record by making a new network request.
+
+**Example**
+
+```js
+polly.configure({
+  expiryStrategy: 'error'
 });
 ```
 
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -166,6 +166,13 @@ export default class Adapter {
   async record(pollyRequest) {
     pollyRequest.action = ACTIONS.RECORD;
 
+    if ('navigator' in global && !navigator.onLine) {
+      console.warn(
+        '[Polly] Recording will fail because the browser is offline.\n' +
+          `${stringifyRequest(pollyRequest)}`
+      );
+    }
+
     return this.onRecord(pollyRequest);
   }
 
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -144,7 +144,7 @@ export default class Adapter {
 
     if ('navigator' in global && !navigator.onLine) {
       console.warn(
-        '[Polly] Recording will fail because the browser is offline.\n' +
+        '[Polly] Recording may fail because the browser is offline.\n' +
           `${stringifyRequest(pollyRequest)}`
       );
     }
@@ -162,19 +162,30 @@ export default class Adapter {
       if (isExpired(recordingEntry.startedDateTime, config.expiresIn)) {
         const message =
           'Recording for the following request has expired.\n' +
-          `${recordingEntry.request.method} ${recordingEntry.request.url}\n`;
+          `${stringifyRequest(pollyRequest, null, 2)}`;
 
-        if (config.expiryStrategy === EXPIRY_STRATEGIES.ERROR) {
-          assert(message);
-        } else if (config.expiryStrategy === EXPIRY_STRATEGIES.WARN) {
-          console.warn(`[Polly] ${message}`, recordingEntry);
-        } else if (config.expiryStrategy === EXPIRY_STRATEGIES.RECORD) {
-          return this.record(pollyRequest);
-        } else {
-          // TODO should there be a generic validator available for all config options?
-          assert(
-            `Invalid config option passed for "expiryStrategy": "${config.expiryStrategy}"`
-          );
+        switch (config.expiryStrategy) {
+          // exit into the record flow if expiryStrategy is "record".
+          case EXPIRY_STRATEGIES.RECORD:
+            console.log('using record strategy');
+
+            return this.record(pollyRequest);
+          // throw an error and exit if expiryStrategy is "error".
+          case EXPIRY_STRATEGIES.ERROR:
+            console.log('using error strategy');
+            this.assert(message);
+            break;
+          // log a warning and continue if expiryStrategy is "warn".
+          case EXPIRY_STRATEGIES.WARN:
+            console.log('using warn strategy');
+            console.warn(`[Polly] ${message}`);
+            break;
+          // throw an error if we encounter an unsupported expiryStrategy.
+          default:
+            this.assert(
+              `Invalid config option passed for "expiryStrategy": "${config.expiryStrategy}"`
+            );
+            break;
         }
       }
 
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -167,17 +167,13 @@ export default class Adapter {
         switch (config.expiryStrategy) {
           // exit into the record flow if expiryStrategy is "record".
           case EXPIRY_STRATEGIES.RECORD:
-            console.log('using record strategy');
-
             return this.record(pollyRequest);
           // throw an error and exit if expiryStrategy is "error".
           case EXPIRY_STRATEGIES.ERROR:
-            console.log('using error strategy');
             this.assert(message);
             break;
           // log a warning and continue if expiryStrategy is "warn".
           case EXPIRY_STRATEGIES.WARN:
-            console.log('using warn strategy');
             console.warn(`[Polly] ${message}`);
             break;
           // throw an error if we encounter an unsupported expiryStrategy.
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -1,4 +1,10 @@
-import { ACTIONS, MODES, Serializers, assert } from '@pollyjs/utils';
+import {
+  ACTIONS,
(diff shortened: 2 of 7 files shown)


docs/configuration.md
```diff
@@ -54,7 +54,7 @@ polly.configure({
 });
 ```
 
-## recordIfExpired
+## recordIfExpired - _deprecated! use expiryStrategy_
```
sankaritan: I think we forgot to push the latest update to this file. This one breaks formatting a bit..

packages/@pollyjs/adapter/src/index.js
```diff
@@ -52,36 +58,6 @@ export default class Adapter {
...
-
-        return false;
-      }
-
-      if ('navigator' in global && !navigator.onLine) {
-        console.warn(
-          '[Polly] Recording for the following request has expired but the browser is offline.\n' +
-            `${recordingEntry.request.method} ${recordingEntry.request.url}\n`,
-          recordingEntry
-        );
```
fastfrwrd: Note: we removed this part of the logic in this PR. If we were to include it, we would only need to include it for `expiryStrategy: 'record'`, and we think that it makes more sense to move between expiryStrategies (from record to warn, for example) for this offline case. If there's a case we're not thinking of, let us know and we'll update it. The branching got pretty hard to understand if this were just nested as-is in the `record` case.

fastfrwrd: We're also unsure why this logic would only apply to the expired case and not other "background record" cases, such as `recordIfMissing`.

agraves: Just to clarify, we're referencing the `navigator.onLine` behavior, not recording in general :)

offirgolan: > We're also unsure why this logic would only apply to the expired case and not other "background record" cases, such as recordIfMissing.

I totally agree with you here. I think instead of removing this, it should be moved directly into the `record` method.

packages/@pollyjs/adapter/src/index.js
```diff
@@ -176,8 +152,23 @@ export default class Adapter {
     if (recordingEntry) {
       await pollyRequest._emit('beforeReplay', recordingEntry);
 
-      if (this.shouldReRecord(pollyRequest, recordingEntry)) {
-        return this.record(pollyRequest);
+      if (isExpired(recordingEntry.startedDateTime, config.expiresIn)) {
+        const message =
+          'Recording for the following request has expired.\n' +
+          `${recordingEntry.request.method} ${recordingEntry.request.url}\n`;
```
offirgolan: You can use `stringifyRequest(pollyRequest)` utility method here instead.

packages/@pollyjs/core/src/utils/merge-configs.js
```diff
@@ -1,4 +1,32 @@
 import mergeWith from 'lodash-es/mergeWith';
+import { assert, EXPIRY_STRATEGIES } from '@pollyjs/utils';
+
+function deprecateRecordIfExpired(mergedConfig) {
+  const bothAreUsed =
+    mergedConfig.hasOwnProperty('recordIfExpired') &&
+    mergedConfig.hasOwnProperty('expiryStrategy');
```
offirgolan: This will always be `true` since we merge the config with all the defaults.

packages/@pollyjs/core/src/defaults/config.js
```diff
@@ -16,10 +16,10 @@ export default {
   logging: false,
 
   recordIfMissing: true,
-  recordIfExpired: false,
   recordFailedRequests: false,
 
   expiresIn: null,
+  expiryStrategy: 'replay',
```
offirgolan: This should be defaulted to `warn`.

packages/@pollyjs/core/src/utils/merge-configs.js
```diff
@@ -1,4 +1,32 @@
...
+
+  if (mergedConfig.recordIfExpired) {
+    // replace recordIfExpired: true with expiryStrategy: record
+    mergedConfig.expiryStrategy = EXPIRY_STRATEGIES.RECORD;
+  } else {
+    // replace recordIfExpired: false with expiryStrategy: warn
+    mergedConfig.expiryStrategy = EXPIRY_STRATEGIES.WARN;
+  }
+
+  return mergedConfig;
```
offirgolan: We should delete the `recordIfExpired` option at the end here. No need to keep it around 😄

packages/@pollyjs/core/src/defaults/config.js
jasonmit: Might also be worth using the `EXPIRY_STRATEGIES` import

packages/@pollyjs/adapter/src/index.js
```diff
@@ -176,8 +152,23 @@ export default class Adapter {
...
+
+        if (config.expiryStrategy === EXPIRY_STRATEGIES.ERROR) {
+          this.assert(message);
+        } else if (config.expiryStrategy === EXPIRY_STRATEGIES.WARN) {
+          console.warn(`[Polly] ${message}`, recordingEntry);
+        } else if (config.expiryStrategy === EXPIRY_STRATEGIES.RECORD) {
+          return this.record(pollyRequest);
+        } else {
+          // TODO should there be a generic validator available for all config options?
```
jasonmit: Good call out and makes sense to me. Could be a future PR since it's beyond the scope of this PR

fastfrwrd:
@jasonmit @offirgolan do y'all have a way of iterating on a single test (`it` block) with testem? I didn't see one in the CONTRIBUTING guide, and testem is new to all of us.

offirgolan:
You can run a single test by using it.only on the test and then running `yarn test -l Chrome` or `yarn test:ci -l Chrome`. To run the test in node, substitute `-l Chrome` with `-l Node`. 

I'll make sure to add that to the Contributing guide as you're not the first to ask 😄

tests/integration/adapter-tests.js
```diff
@@ -57,6 +57,90 @@ export default function adapterTests() {
...
+    await new Promise(r => setTimeout(r, 10));
+
+    await this.relativeFetch(url);
+    await persister.persist();
+    const { startedDateTime: newDateTime } = await persister.findEntry(
+      pollyRequest
+    );
+
+    expect(new Date(prevDateTime)).to.be.lessThan(new Date(newDateTime));
+  });
```
fastfrwrd: we're having a pretty hard time figuring out the best way to test expiration. we'd thought that persisting in between checks in the test would clear the persister cache and allow us to test that the file had a timestamp that was then "expired" given a very low expiration time; any help y'all can provide would be really helpful

offirgolan: Ok, took me a little bit of debugging but I figured out what was going on. The persister does reset the cache but the `persister` instance you're holding on to is an old one (its a pretty troll behavior that we can get around by doing diffing of the new and old config to determine what needs to be created). 

A new persister instance was created after you did `this.polly.configure()`. The following should fix it: 

```js
    this.polly.configure({
      recordIfExpired: true,
      expiresIn: '1ms',
      matchRequestsBy: {
        order: false
      }
    });

    const url = '/api';
    const { persister, server } = this.polly;
```

In terms of testing strategy, I think that testing against the startedDateTime should be sufficient but you can also tap into the `beforePersist` event to test that it was being triggered twice.

offirgolan: Just as a note here, you're going to have to manually delete the recording after every test since its not something we want to be persisted on the fs. You can use `await persister.delete(recordingId)` to do so. Might be worth moving all theses expiration tests to be under their own `describe` block.

packages/@pollyjs/core/src/utils/merge-configs.js
jasonmit: I believe the defaults were updated and the prop was removed, so this comment no longer applies.

tests/integration/adapter-tests.js
fastfrwrd: yep, we were thinking that would be needed. We're thinking a better solution for these tests longer-term would be to implement an in-memory persister to use for some tests that test the internals like this; maybe we'd contribute that in the future if there are other uses of an in-memory persister

fastfrwrd:
Ok, so this is quite close now. Writing the tests for this has been pretty harrowing, but it's worth it.

1. I had to [disable the tests for puppeteer and XHR for the thrown error](https://github.com/Netflix/pollyjs/pull/227/files#diff-e3abd03d1ce1b8db8148b753a77377a5R280) because they always fail.
2. When I skip XHR on that test, I somehow have polluted the test running to the point that two XHR Integration tests fail (which we see in the failed tests on CI).

tests/integration/adapter-tests.js
offirgolan: Interesting! An in-memory persister can be quite useful just for testing purposes. I'll open a new issue for it to better track this and see if I can come up with more uses-cases.

tests/integration/adapter-tests.js
```diff
@@ -207,4 +207,97 @@ export default function adapterTests() {
...
+        mode: 'replay',
+        expiresIn: '1ms',
+        matchRequestsBy: {
+          order: false
+        }
+      });
+    });
+
+    afterEach(async function() {
+      await this.polly.stop();
```
offirgolan: I dont think this stop is necessary here as `setupPolly` should already do that.

offirgolan:
@fastfrwrd I'm gonna look into the XHR & Puppeteer issues and try to figure out why they might be failing.

offirgolan:
So the reason as to why that test was failing on some adapters and not others is because each adapter has a different way of handling errors. The XHR adapter for instance does the following:

https://github.com/Netflix/pollyjs/blob/de66b42b8384d7dabba59783c59408dc0053690b/packages/%40pollyjs/adapter-xhr/src/index.js#L61-L66

The solution is to get the error from the `error` event which will guarantee that the error was emitted since we have a top level error handler.

This test should pass across all adapters:

```js
it('throws on expired recording if expiryStrategy is "error"', async function() {
  const { server } = this.polly;
  let error;

  this.polly.configure({ expiryStrategy: 'error' });
  server.any().on('error', (req, e) => (error = e));

  try {
    await testExpiration.call(this);
  } catch (e) {
    // noop
  }

  expect(error).to.exist;
  expect(error.message).to.match(
    /Recording for the following request has expired/
  );
});
```

packages/@pollyjs/core/src/utils/merge-configs.js
fastfrwrd: removed this check, it was happening every time we had both. For now, `recordIfExpired` just overrides `expiryStrategy`, but it logs a console.warn if you use it.

packages/@pollyjs/core/src/utils/merge-configs.js
fastfrwrd: done!

tests/integration/adapter-tests.js
fastfrwrd: removed!

fastfrwrd:
OK!! The tests are done and the comments are addressed, I believe. Let me know if there are any other things we should take a look at! Thank you SO MUCH for your help on the tests @offirgolan!

packages/@pollyjs/adapter/src/index.js
```diff
@@ -176,8 +159,34 @@ export default class Adapter {
...
-        return this.record(pollyRequest);
+      if (isExpired(recordingEntry.startedDateTime, config.expiresIn)) {
+        const message =
+          'Recording for the following request has expired.\n' +
+          `${stringifyRequest(pollyRequest, null, 2)}`;
+
+        switch (config.expiryStrategy) {
+          // exit into the record flow if expiryStrategy is "record".
+          case EXPIRY_STRATEGIES.RECORD:
+            console.log('using record strategy');
```
offirgolan: Remove `console.log`

fastfrwrd: Ok, removed!

offirgolan:
Released with `v2.6.0` 🎉

PR #236, opened 2019-07-18
fix: use watch strategy

### LINKED ISSUE #232: running polly test suite (yarn test) uses a lot of CPU
# Description

I've noticed that running `yarn test` is slow and uses a lot of CPU, making it difficult to run tests in a tight loop when TDDing. As configured, the testem suite runs a full `yarn test:clean` followed by a `yarn build test:build` for each of the submodules in the monorepo.

I've got a '17 mbp, and just running `$ yarn test` puts my load over 200 for a solid couple minutes. I also made the mistake of mashing enter a few times in the testem ui, which wound up kicking off 143 yarn processes making my machine beachball until I gave up. This load is almost entirely due to running a full build and clean before each suite run.

While fully saturating the machine's resources may make sense in e.g. a CI environment, it took us a bit of investigation to figure out how to avoid melting our machines, and we worry that it may be a bit of a deterrent to new contributors.

So, after a lot of investigation, we wound up with this workflow:

1. Comment out the `before_tests` and `on_start` entries in `testem.js`
2. Manually clean & launch the testem watcher (with logging and a single target): `yarn test:clean && yarn testem -d testem.log -l Node`
3. Then, in a second tab, we'd manually trigger the build: `yarn test:clean && ./node_modules/.bin/run-s build test:build`

At this point, we were able to modify tests and the watcher would quickly kick-off a re-run. I'm not sure exactly when it's necessary to re-run the build, but especially when TDDing existing code, we were able to run the tests in a tight loop.

We also seem to be able to just run the build command for a specific sub-module if we're modifying it.

Anyways, tl;dr decoupling the build and test run steps seems to be the difference between a really long and cpu-intensive wait and fairly instantaneous test results.

So —

1. So first, I'd like to know — do you guys encounter this problem? Or are you using a different workflow we're not aware of?
2. If we're not missing something, I'd propose that we document / set up a workflow that either depends on the developer to build as needed, or selectively watches and builds just the changed parts of the app.

I'd be happy to take a swing at either the manual or selective build workflows but thought it would be good to broach the discussion here first!

Thanks for reading 

### Config

I'm just checking out pollyjs' master and running the tests with the default config.

## Environment

OSX, yarn 1.17.3, node 10.16.0

agraves:
NB: Fixes #232

NB: submitting this for a gut-check / as a proof of concept. There's a lot of style / cleanup work that needs to be done before this is ready for a line-by-line.

## Motivation and Context

Let's say you're a new developer working on Polly. You check out the repo, install deps and run `$ yarn test`. This kicks off the following steps:

1. `yarn clean`
1. `yarn build` for server & utils
1. `yarn build` and `test:build` for ~12 sundry sub-packages
1. `yarn testem`, which kicks off the test runner.

This is a development pain point for the following reasons:

1. These tasks are run as a pre-hook inside `testem`, which makes the testing ui unresponsive for the ~2m it take to run the build.
1. Users are forced to re-run the build process whenever they modify either the tests or the application.
1. Hitting enter in the `testem` ui re-triggers the hooks, which can saturate your CPU if you're impatient.

## Description

While this required a lot of trial and error, the changes are actually fairly contained:

1. `testem` no longer runs any hooks. Running long tasks makes the UI unstable.
1. `testem` now only re-runs on changes to build artifacts (code and tests) instead of the source
1. Every sub-package now has a `yarn watch` command that triggers a watch on its code and tests
1. The root `package.json` now has a `yarn watch` command that runs all of the sub-packages' watch commands in parallel.

## Testing instructions

* in one tab, run `yarn watch` and wait for it to settle
* run `yarn test` in another tab
* trigger a failure by, for example, adding "`expect(false).to.equal(true)`" line 6 of `tests/integration/adapter-browser-tests.js`
* observe that polly selectively re-builds the select file and re-triggers the tests

### General concerns
1. I decided to go with this approach because it was the only one I could find that didn't require me to take a chainsaw to the actual configs.
1. The best approach I can find would be to use (https://github.com/rollup/rollup/pull/1389)(https://github.com/rollup/rollup/pull/1389) approach. tl;dr you can pass an array of rollup configs to a single rollup instance. This would allow us to have a single rollup instance watch every sub-package instead of having 15 separate processes, but trying to aggregate the final configs turned out to be really painful (because of liberal use of `process.cwd` among other things).

### Developer workflow
As implemented, this requires developers to manually kick off watch, see when it settles and then start testem. I personally prefer this approach, but I suspect something more streamlined will be required. So, I'd propose the following:

1. `yarn test` should run `yarn clean && yarn build && yarn testem` and then launch `yarn watch` in the background. I'm not exactly sure how this would be implemented, but quitting testem would *have* to terminate the watchers.
1. Developers who want a little more control can use `yarn test:watch` and `yarn build:watch` to use the workflow I'm submitting with -- i.e. complete separation of test running and build, which imo is preferable if for example you are booting testem and know that an incremental build is sufficient. (As an aside, an incremental build is actually *always* sufficient since it can build from scratch -- the problem is that there's no obvious way to block testem until the incremental build has finished, aside from writing something to block that process until every artifact is present).

## Merge blockers

- [ ] style check for commit, code changes and so forth
- [ ] `test/integration/adapter-node-tests.js` seems to not be properly run on master?
- [ ] look into the ember submodule -- seems to be excluded
- [ ] talk to polly team and agree on naming script commands
- [ ] remove redundant script commands

## Types of Changes

## Checklist


tests/integration/adapter-node-tests.js
```diff
@@ -2,6 +2,7 @@
 
 export default function adapterNodeTests() {
   it('should handle recording requests posting a Buffer', async function() {
+    expect(false).to.equal(true);
```
palmerj3: Remove?

agraves: This is demonstrating that this test doesn’t appear to be running (even on master) (see merge blocker list)

jasonmit:
I believe the quickest win is to disable testem's watcher option by default.  I noticed awhile ago that the tests would re-execute midway through a test run and have long suspect it was because of `watch_files` but never went back to verify.

This morning I found when `watch_files` is set, the tests execute twice which tells me perhaps that testem is detecting a fs change midway through the test run causing them to re-execute.  I think a quick win would be to disable `watch_files` until we're able to figure out exactly why the rebuilds are happening.  When it's disabled, I see a noticeable speed improvement and but the cost is the developer experience suffers in that whenever a test or src file changes, the developer needs to hit `[enter]` to trigger the rebuild.  I think personally think it's a fine compromise until we determine what folder/files are being watched that shouldn't be.

agraves:
@jasonmit Hey Jason -- appreciate the response & detective work. I agree that in the very short term, looking at the watch behavior will make the development experience much more pleasant.

However, I'm eager to clean up and merge this PR since using a watch strategy means that for repeated test runs (e.g. when you're TDDing) you only re-build the changed files, cutting even that ~22s down to a second or so.

jasonmit:
@agraves looking forward to seeing this land.  Let me know if there's anything I can help with.

agraves:
@jasonmit ok thanks man! been away at a workshop but will be back on this next week -- excited to get it over the finish line

agraves:
Ok guys -- @jasonmit et al -- think this is ready for review.

As mentioned elsewhere, this supports two workflows. There's the existing`$ yarn test:ci`, which runs the whole suite.

This now adds a second workflow, where you `$ yarn test:watch` in one tab, wait for it to settle and then `$ yarn test` in another tab. Modifying a test or package only triggers an incremental build, making the suite highly responsive.

Personally I like the explicit nature of that workflow (and the simplicity of its implementation) but I'm open to adding other workflows as well.

jasonmit:
@agraves sounds good, I'll pull it down tomorrow and kick the tires.  Thanks again!

jasonmit:
The build experience is much better!

Some things I need we need are a helpful message explaining to run a build/`yarn watch` first before `yarn test`.  `yarn clean && yarn test` to see what I mean.

I also think we need to add a section to the README or contributor section explaining this setup before merging.

agraves:
Yay! Thanks @jasonmit, glad you liked it. It's a big improvement on our end as well.

I'll go ahead and update the documentation, though I probably won't finish it until early next week.

agraves:
@jasonmit Ok, pushing out some fixes for this now. Here are the key changes:

1. Felt it was a little confusing to have `yarn test` (which runs testem) require you to have the watch process running. So, I renamed `yarn test` to `yarn test:watch` and `yarn test:ci` to `yarn test`. This feels much clearer to me -- a new user can run `yarn test` and get the CI red/green, and then when the read the docs, they will learn to use `yarn watch` and `yarn test:watch` to run incremental builds. Not super married to this but feel pretty good.
2. I updated `CONTRIBUTING.md`, which seems to be the only place that dives into the test commands.
3. I added a bash script that checks for the existence of a couple key files to detect if you're trying to run `yarn test:watch` without having run a build. Unfortunately, the standard bash file-exists check doesn't support wildcards, so I just checked for build artifacts generated by core. I could add something that asserts that the wildcards match at least one file but it makes the script a bit more complicated.

agraves:
Think this is all set from my perspective! 🚀

jasonmit:
Thanks for taking this on @agraves 🥂

PR #257, opened 2019-09-18
fix(adapter-node-http): Correctly handle uploading binary data

### LINKED ISSUE #254: Polly.JS doesn't encode multipart/form-data requests properly
## Description

_This is a bug report._

Polly.JS appears to not send all of a `multipart/form-data` request to servers when it captures the initial server response. Most servers return an error when they cannot parse multipart form boundaries in a request.

**Update:** Polly.JS doesn't pass multipart/form-data bodies with the right encoding. See comments below.

### Shareable Source

I created a minimal reproduction here that uses [request](http://github.com/request/request) to upload a file to a simple [express](https://github.com/expressjs/express) server: https://github.com/swashcap/pollyjs-multipart-form-data-error

Most servers return an error when they cannot parse multipart form boundaries in a request. Polly.JS emits an error similar to:

```
PollyError: [Polly] [persister:fs] Cannot persist response for [GET] http://localhost:3001/ because the status code was 400 and `recordFailedRequests` is `false`
```

### Config

I'm using [setup-polly-jest](https://github.com/gribnoysup/setup-polly-jest) to run tests in Node.js. Here's the config:

```javascript
const AdapterNodeHTTP = require('@pollyjs/adapter-node-http')
const PersisterFS = require('@pollyjs/persister-fs')
const path = require('path')
const { Polly } = require('@pollyjs/core')
const { setupPolly } = require('setup-polly-jest')

Polly.register(AdapterNodeHTTP)
Polly.register(PersisterFS)

// ...

describe('this great test', () => {
  const context = setupPolly({
    adapters: ['node-http'],
    persister: 'fs'
  })

  // ...
})
```

(see the test here: https://github.com/swashcap/pollyjs-multipart-form-data-error/blob/master/src/upload.spec.js)

### Dependencies

```shell
$ yarn list --pattern=polly --silent
├─ @pollyjs/adapter-node-http@2.6.2
├─ @pollyjs/adapter@2.6.0
├─ @pollyjs/core@2.6.2
├─ @pollyjs/node-server@2.6.0
├─ @pollyjs/persister-fs@2.6.2
├─ @pollyjs/persister@2.6.2
├─ @pollyjs/utils@2.6.0
└─ setup-polly-jest@0.5.2
```

## Environment

```shell
$ node -e "var os=require('os');console.log('Node.js ' + process.version + '\n' + os.platform() + ' ' + os.release())"
Node.js v10.16.3
darwin 18.7.0
$ npm --version
6.9.0
$ yarn --version
1.17.3
```

swashcap:
## Description

This fixes a `multipart/form-data` request payload bug by correcting the character encoding. According to https://github.com/nock/nock/issues/191#issuecomment-218791862, nock passes a `multipart/form-data` request's body as a `hex` encoded string. `@pollyjs/adapter-node-http` doesn't transform this back to `utf8`, instead passing the `hex` encoded payload through to the upstream service, causing errors.

## Motivation and Context

Please see https://github.com/Netflix/pollyjs/issues/254 for details.

## Types of Changes

## Checklist

Closes #254.


packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
```diff
@@ -109,4 +113,30 @@ function commonTests(transport) {
 
     expect(nativeHash).to.equal(recordedHash);
   });
+
+  it('should be able to upload a file', async function() {
```
swashcap: I'm having a difficult time writing a test for this. Questions:

* Is there a way to run a _single_ test? Following [the recommendations](https://github.com/Netflix/pollyjs/blob/master/CONTRIBUTING.md#running-only-changed-tests) is very slow on my machine, and Testem's output is difficult to parse.
* Is this even the right location for a `multipart/form-data` test?
* How could I test the entire code path? It seems like `nativeRequest` just does some mocking?

packages/@pollyjs/adapter-node-http/src/index.js
```diff
@@ -73,6 +73,14 @@ export default class HttpAdapter extends Adapter {
         // dealing with json content to convert it back to a string.
         if (body && typeof body !== 'string') {
           body = JSON.stringify(body);
+        } else if (
+          req.headers['content-type'] &&
```
swashcap: This interceptor has access to the `content-type` header, so just check it here: if it's a multipart form request, transform the `body`.

packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
```diff
@@ -109,4 +113,30 @@ function commonTests(transport) {
...
+    );
+    const body = await getStream(formData);
+
+    await nativeRequest(transport, url, {
+      body,
+      headers: formData.getHeaders(),
+      method: 'POST'
+    });
+
+    expect(requests.length).toBe(1);
```
swashcap: I'm noticing `requests[0].body` is empty due to https://github.com/Netflix/pollyjs/blob/7174b25ef3e97e07d22abac1d25a0e38eb9585b1/packages/%40pollyjs/adapter-node-http/tests/utils/native-request.js#L16. Is there a better mechanism for testing this?

packages/@pollyjs/adapter-node-http/src/index.js
jasonmit: Should this test that body is a buffer?

packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
jasonmit: > Is there a way to run a single test? Following the recommendations is very slow on my machine, and Testem's output is difficult to parse.

I recommend the setup described under `"Running only changed tests"`.  This does not suffer from the same performance problems that `yarn test` has.

> Is this even the right location for a multipart/form-data test?

Since this test is specific to this adapter, I believe this is the correct place for it.

> How could I test the entire code path? It seems like nativeRequest just does some mocking?

If you want a more e2e/acceptance style test (which I think this may need) you can leverage the express server that runs along side when our tests run:
https://github.com/Netflix/pollyjs/blob/master/tests/middleware.js

packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
jasonmit: It should be deleting it from the internal `opts` object, it shouldn't be empty because of `delete opts.body`.  Should not effect the `req` object you are pushing to the requests array on line 124

packages/@pollyjs/adapter-node-http/src/index.js
swashcap: :+1: Great idea, will update it.

packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
swashcap: Thanks, I'll have a look at both testing options.

packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
jasonmit: Update: fixed typo above, should now make more sense :)

arty-name:
I’ve hit this issue as well, with a slight difference. I had to remove the `.toString()` call on the Buffer because my jpeg file cannot be represented as a string. The Buffer just works. Could we have it like that?

Is there anything else left to do for this PR? The failing tests seem to be unrelated to the changes.

PR #258, opened 2019-09-25
Adds some test coverage around request IDs

jasonmit:
* Makes the browser integration tests identical in behavior across adapters
* Asserts that the browser integration tests equal the same request ID
* Adds static json POST request ID assertion across all adapters


jasonmit:
I'll reopen in a few hours, need to do some more work on it

PR #259, opened 2019-09-26
fix(adapter-fetch): Correctly handle Request instance passed into fetch

### LINKED ISSUE #255: Bug: Fetch Adapter support for Request objects is incomplete
As of https://github.com/Netflix/pollyjs/pull/220, the Fetch Adapter allows you pass in a Request object as the only argument. However, it will simply extract the URL from the Request, and not use any of the other parameters.

This means if I have code being tested that uses Request objects to POST with a bunch of headers, like this:
```
const reqOpts = {
    body: xml,
    method: "POST",
    headers: {
      Accept: "application/xml",
      "Content-Type": "application/xml",
      "X-custom-header": "secret value"
    }
  };
req = new Request(url, reqOpts);
fetch(req);
```

Polly will do a GET to `url`, with a minimal set of headers.

I suspect this could be mostly fixed by modifying this code:
https://github.com/Netflix/pollyjs/blob/master/packages/%40pollyjs/adapter-fetch/src/index.js#L41

To be:
```
      // Support Request object
      if (typeof url === 'object' && 'url' in url) {
        options = url;
        url = url.url; 
      }
```

I'm testing this solution for us, I'll let you know how it goes.

offirgolan:
## Description

Correctly handle passing a Request instance to fetch. 

```js
fetch(new Request('/api/foo', { method: 'POST' }), { body: 'bar' });
```

## Motivation and Context

Resolves #255.

The approach taken to solve for this is by patching the global Request class and storing the passed in `url` and `options`. This allows us to not only have direct access to the passed in body (e.g. no need to re-read it off the Request instance) but also not needing to hardcode which properties to take off the instance. 

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/adapter-fetch/src/index.js
+++ b/packages/@pollyjs/adapter-fetch/src/index.js
@@ -5,7 +5,6 @@ import serializeHeaders from './utils/serializer-headers';
 
 const { defineProperty } = Object;
 const IS_STUBBED = Symbol();
-const IS_REQUEST = Symbol();
 const REQUEST_ARGUMENTS = Symbol();
 
 export default class FetchAdapter extends Adapter {
@@ -45,7 +44,7 @@ export default class FetchAdapter extends Adapter {
       `await req.blob()` as well as not having to hard code each option we want
       to extract from the Request instance.
     */
-    context.Request = class Request extends context.Request {
+    class ExtendedRequest extends context.Request {
       constructor(url, options) {
         super(url, options);
 
@@ -58,7 +57,7 @@ export default class FetchAdapter extends Adapter {
           the first argument so we use its arguments and merge it with the
           new options.
         */
-        if (typeof url === 'object' && url[IS_REQUEST]) {
+        if (url instanceof ExtendedRequest) {
           const reqArgs = url[REQUEST_ARGUMENTS];
 
           args = { ...reqArgs, options: { ...reqArgs.options, ...options } };
@@ -66,22 +65,22 @@ export default class FetchAdapter extends Adapter {
           args = { url, options };
         }
 
-        defineProperty(this, IS_REQUEST, { value: true });
         defineProperty(this, REQUEST_ARGUMENTS, { value: args });
       }
 
       clone() {
-        return new Request(this);
+        return new ExtendedRequest(this);
       }
-    };
+    }
 
+    context.Request = ExtendedRequest;
     defineProperty(context.Request, IS_STUBBED, { value: true });
 
     context.fetch = (url, options = {}) => {
       let respond;
 
       // Support Request object
-      if (typeof url === 'object' && url[IS_REQUEST]) {
+      if (url instanceof ExtendedRequest) {
         const req = url;
         const reqArgs = req[REQUEST_ARGUMENTS];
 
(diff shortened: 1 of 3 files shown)


packages/@pollyjs/adapter-fetch/tests/integration/adapter-test.js
```diff
@@ -75,8 +75,8 @@ describe('Integration | Fetch Adapter | Init', function() {
         adapterOptions
       });
 
-      expect(polly.adapters.get('fetch').native).to.equal(fetch);
-      expect(polly.adapters.get('fetch').native).to.not.equal(
+      expect(polly.adapters.get('fetch').nativeFetch).to.equal(fetch);
```
jasonmit: Should add tests around this new Request stub

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -26,23 +28,72 @@ export default class FetchAdapter extends Adapter {
...
+
+    defineProperty(context.Request, IS_STUBBED, { value: true });
 
     context.fetch = (url, options = {}) => {
+      let respond;
+
       // Support Request object
-      if (typeof url === 'object' && 'url' in url) {
-        url = url.url;
+      if (typeof url === 'object' && url[IS_REQUEST]) {
```
jasonmit: Could use `instanceof Request` here instead and would eliminate the additional `IS_REQUEST` prop

offirgolan: That would break if an actual native request gets passed in since we extend the native Request class.

```js
class R extends window.Request {}

new R('/') instanceof window.Request // true
new R('/') instanceof R // true
```

jasonmit: Can compare `instanceOfR.constructor === R`

jasonmit: Actually I'm not sure I understand your comment.

Are you saying you worry that an instance of `Request,` that isn't an instance of our `ExtendedRequest` class and instead the native `Request`, would get through an `instanceof ExtendedRequest` check?  That shouldn't be the case if so.

offirgolan: Ah, yeah you're totally right. My mistake above is not always comparing to the ExtendedRequest 🤦‍♂. I'll make the change to use `instanceof` 😄

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -68,14 +118,20 @@ export default class FetchAdapter extends Adapter {
...
+    this.nativeFetch = null;
+    this.NativeRequest = null;
   }
 
   async passthroughRequest(pollyRequest) {
+    const { context } = this.options;
     const { options } = pollyRequest.requestArguments;
 
-    const response = await this.native.apply(global, [
+    const response = await this.nativeFetch.apply(context, [
```
jasonmit: Nice catch

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -26,23 +27,72 @@ export default class FetchAdapter extends Adapter {
...
+    this.nativeFetch = context.fetch;
+    this.NativeRequest = context.Request;
+
+    /*
+      Patch the Request class so we can store all the passed in options. This
+      allows us the access the `body` directly instead of having to do
+      `await req.blob()` as well as not having to hard code each option we want
+      to extract from the Request instance.
+    */
+    class ExtendedRequest extends context.Request {
```
jasonmit: nit: `ExtendedRequest` can be hoisted up and out of `onConnect()`

Issue #244, opened 2019-08-01
Snapshot regression tests


offirgolan:
## Description

Because of the use of the InMemoryPersister (#233), we removed all the stored recordings generated by the tests which we used as a crutch to do semi-manual regression testing (e.g. check if certain attributes changed in the HAR file after making a change). Instead, we should write actual snapshot tests to catch these which will be significantly more useful.

PR #260, opened 2019-10-01
chore: Adds identifier assertions for xhr and fetch adapters

jasonmit:
Partially solves #244 

Puppeteer I'll tackle in a follow up PR.

### DIFF
--- a/tests/integration/adapter-identifier-tests.js
+++ b/tests/integration/adapter-identifier-tests.js
@@ -1,0 +1,272 @@
+import set from 'lodash-es/set';
+import deepmerge from 'deepmerge';
+
+function adapterIdentifierTests() {
+  describe('matchRequestsBy', function() {
+    beforeEach(function() {
+      const { polly } = this;
+
+      polly.server.post('/*').intercept((req, res) => {
+        res.sendStatus(200);
+      });
+
+      this.requests = captureRequests(polly.server);
+    });
+
+    testConfiguration('method', false, {
+      expected: {
+        id: 'e7e58325a54088ab228cd9dbe7558141',
+        identifiers: {
+          headers: { 'content-type': 'application/json;charset=utf-8' },
+          body: '{}',
+          url: 'http://localhost:4000/pathname?query=param'
+        }
+      },
+      overrides: {
+        'node-http': {
+          id: '815b91955669771aa83a74ca27f6fbd8',
+          identifiers: {
+            headers: {
+              host: 'localhost:4000'
+            }
+          }
+        }
+      }
+    });
+
+    testConfiguration('headers', false, {
+      expected: {
+        id: '74e48a0f5e33321aa1aa3f4f65c8ccde',
+        identifiers: {
+          method: 'POST',
+          body: '{}',
+          url: 'http://localhost:4000/pathname?query=param'
+        }
+      }
+    });
+
+    testConfiguration('body', false, {
+      expected: {
+        id: 'b0427f53912c03a68ea2d4e923e136a8',
+        identifiers: {
+          body: undefined,
+          headers: { 'content-type': 'application/json;charset=utf-8' },
+          method: 'POST',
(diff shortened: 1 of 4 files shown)


tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,308 @@
...
+      });
+
+      testConfiguration('url.username', false, {
+        fetch: {
+          id: '3dbff3c3fbccd6f97d01be31fc7fdd59',
+          identifiers: {
+            headers: { 'content-type': 'application/json;charset=utf-8' },
+            method: 'POST',
+            body: '{}',
+            url: 'http://localhost:4000/pathname?query=param'
```
jasonmit: Looks like we have a bug in Polly, url here dropped `password` from the matchBy.  Here we only turned off username, but password went with it.

tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,308 @@
...
+          }
+        },
+        xhr: {
+          id: '3dbff3c3fbccd6f97d01be31fc7fdd59',
+          identifiers: {
+            headers: { 'content-type': 'application/json;charset=utf-8' },
+            method: 'POST',
+            body: '{}',
+            url: 'http://localhost:4000/pathname?query=param'
```
jasonmit: Same as above

tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,308 @@
...
+          body: undefined,
+          headers: { 'content-type': 'application/json;charset=utf-8' },
+          method: 'POST',
+          url: 'http://a:b@localhost:4000/pathname?query=param'
+        }
+      },
+      xhr: {
+        id: '3dbfc69c4a73f3f368ae00aee80d2379',
+        identifiers: {
+          body: undefined,
```
jasonmit: I don't think `body` should be set on the Map but not a big deal

tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,308 @@
+import set from 'lodash-es/set';
+
+function adapterIdentifierTests() {
```
jasonmit: The general id here is if we change any of the identifiers, we either introduced a breaking change or a bug.  In writing these tests, I did spot existing bug - there could be more.

tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,307 @@
...
+  describe.only('matchRequestsBy', function() {
```
jasonmit: Remove once I'm done with all of this

tests/integration/adapter-identifier-tests.js
```diff
@@ -0,0 +1,307 @@
...
+          identifiers: {
+            headers: { 'content-type': 'application/json;charset=utf-8' },
+            method: 'POST',
+            body: '{}',
+            url: 'http://a:b@localhost/pathname?query=param'
+          }
+        },
+        xhr: {
+          id: '147ca2615501f226981619c32341699b',
+          identifiers: {
```
jasonmit: Decided to assert also against the identifiers, will help when we need to triage why an ID changed.

tests/integration/adapter-identifier-tests.js
offirgolan: Good catch. We shoud create a new issue with this and follow up on this unless you have the time to fix it with this PR.

offirgolan:
@jasonmit this looks great! Thanks for taking this up 😄.

jasonmit:
Awesome, I'll wrap this up over the weekend!

tests/integration/adapter-identifier-tests.js
jasonmit: Looks like the spec doesn't support passing credentials in the URL - results in a TypeError.  I think we should deprecate and eventually remove url.password and url.username options.

PR #284, opened 2019-12-24
update index.html

moonstash:
Problem
Currently your website is down because some of the resources on unpkg.com are reporting as "Rate exceeded."

Solution
Switch from pointing to resources on unpkg to cloudflare which fixes it.


offirgolan:
Thanks for opening this PR @moonstash. This looks to be an intermittent issue that was resolved yesterday (mjackson/unpkg#237).

PR #291, opened 2020-01-10
fix(core): Disconnect from all adapters when `pause` is called

### LINKED ISSUE #285: Should "_order" still increment after polly.pause?
##  Description

Question: I have a code snippet that does roughly:

```
polly.pause()
// keep making HTTP request until response satisfies a condition
polly.play()
// make same HTTP request
```

I would expect the recording to only contain the last request (after `.play()`) and have the `_order` attribute still be 0, as if the requests between pause and play didn't happen. Instead, I see the recording not having the extra requests (as expected), but `_order` becomes something like 7. What's the correct semantics here?

### Config

```js
{
  adapters: ['node-http'],
  persister: 'fs',
  logging: false,
  recordFailedRequests: true,
  recordIfMissing: true,
  mode: __RECORD_MODE__,
  matchRequestsBy: {
    headers: {
      exclude: ['Authorization'],
    },
  },
}
```

### Dependencies

```json
{
  "@pollyjs/adapter-fetch": "^3.0.0",
  "@pollyjs/adapter-node-http": "^3.0.0",
  "@pollyjs/core": "^3.0.0",
  "@pollyjs/persister-fs": "^3.0.0",
  "@pollyjs/persister-rest": "^3.0.0",
}
```

## Environment

```
darwin 18.7.0
NPM 6.9.0
yarn 1.17.3
```

offirgolan:
## Description

Calling `polly.pause()` will now disconnect from all connected adapters instead of setting the mode to passthrough. Calling `polly.play()` will reconnect to the disconnected adapters before pause was called.

## Motivation and Context

Resolves #285.

## Types of Changes

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --frozen-lockfile --non-interactive
+  - yarn install --no-lockfile --non-interactive
 
 script:
   - commitlint-travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
+  - yarn install --frozen-lockfile --non-interactive
 
 script:
   - commitlint-travis
--- a/packages/@pollyjs/core/tests/unit/polly-test.js
+++ b/packages/@pollyjs/core/tests/unit/polly-test.js
@@ -239,6 +239,21 @@ describe('Unit | Polly', function() {
   describe('API', function() {
     setupPolly();
 
+    class MockAdapterA extends Adapter {
+      static get name() {
+        return 'adapter-a';
+      }
+
+      onConnect() {}
+      onDisconnect() {}
+    }
+
+    class MockAdapterB extends MockAdapterA {
+      static get name() {
+        return 'adapter-b';
+      }
+    }
+
     it('.record()', async function() {
       this.polly.mode = MODES.REPLAY;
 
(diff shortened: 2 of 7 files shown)


packages/@pollyjs/core/tests/unit/polly-test.js
```diff
@@ -256,23 +271,35 @@ describe('Unit | Polly', function() {
     });
 
     it('.pause()', async function() {
```
jasonmit: Should add a test asserting on the behavior of _order when a pause and play occurs with requests in between.  Will prevent a regression but also solidify expected behavior.

jasonmit:
Side note: should note this is a breaking change in the commit msg

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
+  - yarn install --frozen-lockfile --non-interactive
```
jasonmit: I think we want to keep this to ensure polly still works with all the latest deps. Otherwise, we won't pick up on breaks until they're reported.

offirgolan: Agreed. It looks like node [semver v7](https://github.com/npm/node-semver/blob/bb36c98d71d5760d730abba71c68bc324035dd36/package.json#L36) is no longer supporting node 8 which is breaking our [node 8 build](https://travis-ci.org/Netflix/pollyjs/jobs/635842564#L533). 

Node 8 is EOL as of this year, should we drop support as well? I suspect more and more packages will be doing the same.

jasonmit: Yup, agree to drop

offirgolan: Sounds good, I'll address in a follow up PR as this is unrelated.

PR #294, opened 2020-01-13
fix(adapter): Clone the recording entry before mutating it

### LINKED ISSUE #265: BUG: beforePersist isn't fired when persisting some responses
## Description
**What I was trying to accomplish:**
> from #256
> I'd really like to have my recordings be human-readable (and editable), so I'd like to find a way to decode content as it's persisted, then re-encode it when it's read.

@jasonmit suggested using `beforePersist` and `beforeReplay` to handle the decoding and encoding, respectively.

**What went wrong:**
> from [original discussion](https://github.com/Netflix/pollyjs/issues/256#issuecomment-540899494)
> it seems that when an existing recording is updated (for example, if I had 5 requests in the recording, but my test scenario ended up recording a new, 6th request), beforePersist does not get triggered for my pre-existing (previously decoded) requests, so when the recording is persisted, it's bypassing my decoding logic, and those pre-existing requests are persisted in encoded, gzipped format (instead of my decoded, human-readable format).

> It seems that beforePersist behaves as more of a beforeFirstPersist, which would be fine if you were trying to prune out sensitive info that isn't needed for subsequent requests, but maybe not so good for round-trip processing like I'm aiming for.

n/a

### Source
here's the source of the encoding/decoding handlers, if that helps:
```typescript
import { PollyServer, RouteHandler } from "@pollyjs/core";
import { gunzipSync, gzipSync } from "zlib";

const contentTypeDelimiter = /;\s*/;
const defaultCharset = "utf-8";

interface RecordingHeader {
  name: string;
  value: string;
}

interface RecordingContent {
  mimeType: string;
  text: string;
  size: number;
}

interface RecordingResponse {
  bodySize: number;
  content: RecordingContent;
  headers: RecordingHeader[];
}

type HeaderSpecification = [string, string | ((value: string) => boolean)];

const getEncoding = (content: RecordingContent): BufferEncoding => {
  // this doesn't know how to deal with multipart media types
  const [, charset = defaultCharset] = content.mimeType.split(
    contentTypeDelimiter
  );

  if (!Buffer.isEncoding(charset)) {
    throw new Error(`'${charset}' is not a valid BufferEncoding value`);
  }

  return charset;
};

const hasHeader = (headers: RecordingHeader[]) => ([
  key,
  value
]: HeaderSpecification): boolean =>
  headers.some(header =>
    header.name === key && typeof value === "function"
      ? value(header.value)
      : header.value === value
  );

const headerValueDelimiter = /,\s*/;
const headerValueContains = (valueToFind: string) => (
  headerValue: string
): boolean => headerValue.split(headerValueDelimiter).includes(valueToFind);

const isContentChunked = (headers: RecordingHeader[]): boolean =>
  hasHeader(headers)(["transfer-encoding", headerValueContains("chunked")]);

const isContentGzipped = (headers: RecordingHeader[]): boolean =>
  hasHeader(headers)(["content-encoding", headerValueContains("gzip")]);

const decodeContent = (response: RecordingResponse): void => {
  const { content, headers } = response;
  if (!isContentGzipped(headers) || content.size === 0) {
    return;
  }

  const parsedBody = JSON.parse(content.text);

  const wholeBody = isContentChunked(headers)
    ? parsedBody.join("")
    : parsedBody;

  const zipped = Buffer.from(wholeBody, "hex");
  const unzipped = gunzipSync(zipped);

  const encoding = getEncoding(content);
  content.text = unzipped.toString(encoding);
};

const encodeContent = (response: RecordingResponse): void => {
  const { content, headers } = response;
  if (!isContentGzipped(headers)) {
    return;
  }

  const encoding = getEncoding(content);

  const unzipped = Buffer.from(content.text, encoding);
  const zipped = gzipSync(unzipped);
  const hex = zipped.toString("hex");
  const body = isContentChunked(headers)
    // re-chunk the content
    ? [hex.slice(0, 20), hex.substring(20)] // making the first chunk 20 chars, as it seems to match polly's typical result
    : hex;
  const text = JSON.stringify(body);
  content.text = text;

  // recalculating the content size in case manual editing of data changes the size
  content.size = text.length;
  response.bodySize = text.length;
};

export const useResponseDecoding = (server: PollyServer): RouteHandler =>
  server
    .any()
    .on("beforePersist", (req, { response }) => decodeContent(response))
    .on("beforeReplay", (req, { response }) => encodeContent(response));

```

### Config

Copy the config used to setup the Polly instance:

```js
const polly = new Polly(myRecordingName, {
  adapters: ["node-http"],
  persister: "fs",
  persisterOptions: {
    fs: {
      recordingsDir: myRecordingsDir
    }
  }
});
```

### Dependencies
```json
{
  "@pollyjs/adapter-node-http": "^2.6.3",
  "@pollyjs/core": "^2.6.3",
  "@pollyjs/persister-fs": "^2.6.3"
}
```

## Environment
`Node.js v10.16.3`
`darwin 18.7.0` (MacOS)
`npm 6.9.0`

offirgolan:
## Description

Clone the recording entry before the `beforeReplay` event so any changes will not actually persist to the stored recording.

## Motivation and Context

Resolves #265.

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -153,7 +153,7 @@ export default class Adapter {
 
   async replay(pollyRequest) {
     const { config } = pollyRequest;
-    let recordingEntry = await this.persister.findEntry(pollyRequest);
+    const recordingEntry = await this.persister.findEntry(pollyRequest);
 
     if (recordingEntry) {
       /*
@@ -163,11 +163,11 @@ export default class Adapter {
         Note: Using JSON.parse/stringify instead of lodash/cloneDeep since
               the recording entry is stored as json.
       */
-      recordingEntry = JSON.parse(JSON.stringify(recordingEntry));
+      const clonedRecordingEntry = JSON.parse(JSON.stringify(recordingEntry));
 
-      await pollyRequest._emit('beforeReplay', recordingEntry);
+      await pollyRequest._emit('beforeReplay', clonedRecordingEntry);
 
-      if (isExpired(recordingEntry.startedDateTime, config.expiresIn)) {
+      if (isExpired(clonedRecordingEntry.startedDateTime, config.expiresIn)) {
         const message =
           'Recording for the following request has expired.\n' +
           `${stringifyRequest(pollyRequest, null, 2)}`;
@@ -193,13 +193,13 @@ export default class Adapter {
         }
       }
 
-      await this.timeout(pollyRequest, recordingEntry);
+      await this.timeout(pollyRequest, clonedRecordingEntry);
       pollyRequest.action = ACTIONS.REPLAY;
 
       return this.onReplay(
         pollyRequest,
-        normalizeRecordedResponse(recordingEntry.response),
-        recordingEntry
+        normalizeRecordedResponse(clonedRecordingEntry.response),
+        clonedRecordingEntry
       );
     }
 
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -153,9 +153,18 @@ export default class Adapter {
 
   async replay(pollyRequest) {
     const { config } = pollyRequest;
-    const recordingEntry = await this.persister.findEntry(pollyRequest);
+    let recordingEntry = await this.persister.findEntry(pollyRequest);
 
     if (recordingEntry) {
+      /*
+        Clone the recording entry so any changes will not actually persist to
+        the stored recording.
+
+        Note: Using JSON.parse/stringify instead of lodash/cloneDeep since
+              the recording entry is stored as json.
+      */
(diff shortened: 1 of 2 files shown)


packages/@pollyjs/adapter/src/index.js
```diff
@@ -153,9 +153,18 @@ export default class Adapter {
...
 
     if (recordingEntry) {
+      /*
+        Clone the recording entry so any changes will not actually persist to
+        the stored recording.
+
+        Note: Using JSON.parse/stringify instead of lodash/cloneDeep since
+              the recording entry is stored as json.
+      */
+      recordingEntry = JSON.parse(JSON.stringify(recordingEntry));
```
jasonmit: nit: create a new var `clonedRecordingEntry` instead of reassigning

Issue #297, opened 2020-01-23
@pollyjs/ember stores recordings in [projectRoot]node_modules/@pollyjs/ember/recordings


nfc036:
`@pollyjs/ember@4.0.0` is storing the recordings in `[projectRoot]/node_modules/@pollyjs/ember/recordings`.
Also, configuration options in config/polly.js are ignored.

It seems to me, that this.root in https://github.com/Netflix/pollyjs/blob/6423839d9aeb9b0318ae69251d8bafcd28933408/packages/%40pollyjs/ember/index.js#L78 and https://github.com/Netflix/pollyjs/blob/6423839d9aeb9b0318ae69251d8bafcd28933408/packages/%40pollyjs/ember/index.js#L86-L89 is pointing to the root of the addon, not to the root of the project using @pollyjs/ember. When I change `this.root` to `this.project.root` everything is working as expected.

I'm using ember-cli@3.15.1.

offirgolan:
Fix released in `v4.0.1`.

Issue #276, opened 2019-11-15
pollyjs config is undefined when reusing ember build and running ember test


shankarsridhar:
Hello pollyjs contributors and devs,
I am not sure if this is a bug report or a question :)

## TL;DR
pollyjs config from consuming ember app is not available inside the polly js addon methods, when reusing the build, which happens when running the command,
 `ember test --path <build folder name>`

## Description
I am using pollyjs/@ember package for my ember project, and it works just fine, when I run the tests using the command, `ember test`. 

But when I try to reuse an existing build of my ember project(`ember test --path <build folder name. by default, 'dist'>`), I run into this error in command line:

```
Cannot read property 'enabled' of undefined

Stack Trace and Error Report: /var/folders/l3/zfkk52zd2qsdh8lbhdz6gy1r000lf7/T/error.dump.466e5463069e16ef55ccda6970db9a12.log
```

On further looking into the logs and debugging the ember node process, the error was happening in the `testemMiddleware()` method in the [`@pollyjs/ember/index.js`](https://github.com/Netflix/pollyjs/blob/ee25c77f1f469f20754b458b582319947f727c7b/packages/%40pollyjs/ember/index.js#L44) file as show below:

```js
testemMiddleware(app) {
    if (this._config.enabled) { // <== `this._config` is undefined here
      registerExpressAPI(app, this._config.server);
    }
  },
```

And the root cause seems to be that pollyjs `this.app.options.pollyjs` config defined and retrieved from the consuming ember app into the addon, is available only in the `included()` hook; but not available inside the other methods.

### Error Message & Stack Trace

```
stack: TypeError: Cannot read property 'enabled' of undefined
    at Class.testemMiddleware (/Users/shsridha/workspace/srd-backstop/gryphon-fe/node_modules/@pollyjs/ember/index.js:45:22)
    at middlewares.forEach.middleware (/Users/shsridha/workspace/srd-backstop/gryphon-fe/node_modules/testem/lib/server/index.js:183:9)
    at Array.forEach (<anonymous>)
    at Server.injectMiddleware (/Users/shsridha/workspace/srd-backstop/gryphon-fe/node_modules/testem/lib/server/index.js:182:19)
    at Server.createExpress (/Users/shsridha/workspace/srd-backstop/gryphon-fe/node_modules/testem/lib/server/index.js:133:10)
    at Server.start (/Users/shsridha/workspace/srd-backstop/gryphon-fe/node_modules/testem/lib/server/index.js:41:10)
[...]

PR #298, opened 2020-01-23
[@pollyjs/ember] Fix root path for config and recordings

nfc036:
See https://github.com/Netflix/pollyjs/issues/297#issue-554225908 for a description of the problem.

### DIFF
--- a/packages/@pollyjs/ember/index.js
+++ b/packages/@pollyjs/ember/index.js
@@ -75,7 +75,7 @@ module.exports = {
       server: {}
     };
 
-    let configPath = path.join(this.root, 'config', 'polly.js');
+    let configPath = path.join(this.project.root, 'config', 'polly.js');
 
     if (fs.existsSync(configPath)) {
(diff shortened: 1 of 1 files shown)


jasonmit:
FYI This will reintroduce https://github.com/Netflix/pollyjs/issues/276

nfc036:
Based on https://ember-cli.com/api/classes/Addon.html#props_root this.root is the root of the addon (./node_modules/@pollyjs/ember). Currently the addon is looking for config/polly.js there (in its own root) and it is storing the recordings there too. 

Wasn't #276 related to the use of the included() function? This function is not re-introduced by the supposed change.

jasonmit:
`this.project` is undefined if you run the scenario in #276 on top of your changes.

jasonmit:
https://github.com/Netflix/pollyjs/pull/299 should resolve this without reintroducing #276

Thanks for bringing this to our attention.

PR #299, opened 2020-01-24
fix(ember): Config read from project root

### LINKED ISSUE #297: @pollyjs/ember stores recordings in [projectRoot]node_modules/@pollyjs/ember/recordings

jasonmit:
`this.root` points to the addon's location within `node_modules` - not the project root.

Since we cannot bank on `this.project` to always be available to find the project root folder (see: #276) we have to use the current working directory of the process to determine the root directory of the project - which Ember also seems to use internally.

Fixes #297

### DIFF
--- a/packages/@pollyjs/ember/index.js
+++ b/packages/@pollyjs/ember/index.js
@@ -7,6 +7,8 @@ const path = require('path');
 const { registerExpressAPI, Defaults } = require('@pollyjs/node-server');
 const parseArgs = require('minimist');
 
+const root = process.cwd();
+
 function determineEnv() {
   if (process.env.EMBER_ENV) {
     return process.env.EMBER_ENV;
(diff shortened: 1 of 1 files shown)


packages/@pollyjs/ember/index.js
```diff
@@ -75,7 +77,7 @@ module.exports = {
       server: {}
     };
 
-    let configPath = path.join(this.root, 'config', 'polly.js');
+    let configPath = path.join(root, 'config', 'polly.js');
```
offirgolan: What are your thoughts on doing something like

`const projectRoot = this.project && this.project.root ? this.project.root : root`?

jasonmit: I like it, will update

jasonmit: Updated

PR #303, opened 2020-01-29
Fix parsing JSON hexadecimal strings

brandon-leapyear:
## Description

Fix parsing JSON strings that look like hexadecimal Buffer streams

## Motivation and Context

I was having an issue where Polly was intercepting my JSON payload, the string `"AdministerDb"`, and turning it into a Buffer `<Buffer ad>`. The problem is that `Buffer.from('AdministerDb', 'hex')` parses as much hex as possible, in this case `Ad`, and treating it as a hexadecimal Buffer stream.

To fix this, I made sure to handle JSON values before handling buffer. I'm making the assumption that no one would be sending binary data with `Content-Type: application/json`, but I think it's a valid assumption. I also added tests and found that the JSON value `false` wasn't being stringified either.

## Types of Changes

## Checklist

- [ ] I have read the [contributing guidelines](https://github.com/Netflix/pollyjs/blob/master/CONTRIBUTING.md).

### DIFF
--- a/packages/@pollyjs/adapter-node-http/src/index.js
+++ b/packages/@pollyjs/adapter-node-http/src/index.js
@@ -7,14 +7,14 @@ import nock from 'nock';
 import {
   normalizeClientRequestArgs,
   isUtf8Representable,
-  isContentEncoded,
-  isJSONContent
+  isContentEncoded
 } from 'nock/lib/common';
 import Adapter from '@pollyjs/adapter';
 import { HTTP_METHODS } from '@pollyjs/utils';
 
 import getUrlFromOptions from './utils/get-url-from-options';
 import mergeChunks from './utils/merge-chunks';
+import { parseBody } from './utils/request-body';
 import urlToOptions from './utils/url-to-options';
 
 const IS_STUBBED = Symbol();
@@ -71,20 +71,7 @@ export default class HttpAdapter extends Adapter {
         );
         const url = getUrlFromOptions(parsedArguments.options);
 
-        if (body) {
-          if (
-            typeof body === 'string' &&
-            !isUtf8Representable(Buffer.from(body, 'hex'))
-          ) {
-            // Nock internally converts a binary buffer into its hexadecimal
-            // representation so convert it back to a buffer.
-            body = Buffer.from(body, 'hex');
-          } else if (isJSONContent(headers)) {
-            // Nock will parse json content into an object. We have our own way
-            // of dealing with json content so convert it back to a string.
-            body = JSON.stringify(body);
-          }
-        }
+        body = parseBody(body, headers);
 
         adapter.handleRequest({
           url,
--- a/packages/@pollyjs/adapter-node-http/src/utils/request-body.js
+++ b/packages/@pollyjs/adapter-node-http/src/utils/request-body.js
@@ -2,17 +2,17 @@ import { isUtf8Representable, isJSONContent } from 'nock/lib/common';
 
 export function parseBody(body, headers) {
   if (body) {
-    if (
+    if (isJSONContent(headers)) {
+      // Nock will parse json content into an object. We have our own way
+      // of dealing with json content so convert it back to a string.
+      return JSON.stringify(body);
+    } else if (
       typeof body === 'string' &&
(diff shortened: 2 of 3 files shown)


offirgolan:
@brandon-leapyear thank you for taking the time in opening this PR! 

> I made sure to handle JSON values before handling buffer. I'm making the assumption that no one would be sending binary data with Content-Type: application/json, but I think it's a valid assumption. 

The request can still be compressed even though the content type is set to `application/json`. I think the underlying issue here is that instead of checking if the string is non-utf8, we should check if its a hex string.

```js
function parseBody(body, headers) {
  if (body === null || body === undefined) {
    return body;
  }

  if (typeof body === 'string') {
    // Check if the body is a hex string
    if (body === Buffer.from(body, 'hex').toString('hex')) {
      // Nock automatically converts a binary buffer into its hexadecimal
      // representation so convert it back to a buffer.
      return Buffer.from(body, 'hex');
    }

    return body;
  }

  if (isJSONContent(headers)) {
    // Nock automatically parses json content, but we have our own way
    // of dealing with json content, so convert it back to a string.
    return JSON.stringify(body);
  }

  return body;
}
```

> I also added tests and found that the JSON value false wasn't being stringified either.

Thank you for catching this and adding test cases!

brandon-leapyear:
@offirgolan I don't know if that completely solves the issue; what if someone sends the JSON string `"abcd"`? That's a valid usecase, and also satisfies `body === Buffer.from(body, 'hex').toString('hex')`.

Additionally, `isUtf8Representable(Buffer.from('abcd', 'hex'))` is also false

brandon-leapyear:
Can you point me to where in Nock a hex string may be passed as the body?

Our callback is specified with
```js
interceptor
  .intercept(...)
  .reply(function(...) { ... })
```

which makes me think that we only hit this branch: https://github.com/nock/nock/blob/master/lib/interceptor.js#L113-L122. The important line here is
```js
this.fullReplyFunction = statusCode
```

`this.fullReplyFunction` is used here: https://github.com/nock/nock/blob/c7363e56ee866635514e5aaa0319232b6edd1b61/lib/playback_interceptor.js#L136-L148

which calls `parseJSONRequestBody` here: https://github.com/nock/nock/blob/c7363e56ee866635514e5aaa0319232b6edd1b61/lib/playback_interceptor.js#L9-L18

which returns a JSON-parsed value (after possibly deflating the body). If this is the case, shouldn't be able to just do

```js
function parseBody(body, headers) {
  if (body === null || body === undefined) {
    return body;
  }

  if (isJSONContent(headers)) {
    // Nock automatically parses json content, but we have our own way
    // of dealing with json content, so convert it back to a string.
    return JSON.stringify(body);
  }

  return body;
}
```

offirgolan:
Yeah, that's actually what it was initially and was changed by #257.

The body is converted to a hex string here:

https://github.com/nock/nock/blob/c7363e56ee866635514e5aaa0319232b6edd1b61/lib/intercepted_request_router.js#L250-L258

Which is then passed into here:

https://github.com/nock/nock/blob/c7363e56ee866635514e5aaa0319232b6edd1b61/lib/playback_interceptor.js#L81-L82

Ideally we would have access to `requestBodyIsUtf8Representable` but instead it looks like we do have access to `requestBodyBuffers` 🎉. That means we can skip all this guessing and hopefully just be able deal with the raw request body buffers.

```js
interceptor.intercept(/.*/, m).reply(function(_, _body, respond) {
  const { req, method } = this;
  const { headers } = req;
  const parsedArguments = normalizeClientRequestArgs(
    ...REQUEST_ARGUMENTS.get(req)
  );
  const url = getUrlFromOptions(parsedArguments.options);
  const requestBodyBuffer = Buffer.concat(req.requestBodyBuffers);
  const body = isUtf8Representable(requestBodyBuffer)
    ? requestBodyBuffer.toString('utf8')
    : requestBodyBuffer;

  adapter.handleRequest({
    url,
    method,
    headers,
    body,
    requestArguments: { req, body, respond, parsedArguments }
  });
});
```

brandon-leapyear:
https://github.com/Netflix/pollyjs/pull/304

PR #305, opened 2020-01-30
test: Updates the jest-puppeteer example

### LINKED ISSUE #290: jest-puppeteer example doesn't work as expected
## Description

Bug - The examples folder `jest-puppeteer` doesn't work, i imagine that is a topic of version because all package are configured to install last version "*".
Installed all packages then run `yarn test`

The console doesn't print that the test failed or not, another thing i did for test was delete the recording folder and it doesn't recorded a new one.

I've tried those packages into my local application and they didn't work as well, when trying puppeteer alone, including the examples folder everything works as expected.

jasonmit:
fixes #290

Pretty sure our other examples need some love also, I'll deal with them in a future PR.

### DIFF
--- a/examples/jest-puppeteer/package.json
+++ b/examples/jest-puppeteer/package.json
@@ -8,12 +8,12 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@pollyjs/adapter-puppeteer": "*",
-    "@pollyjs/core": "*",
-    "@pollyjs/persister-fs": "*",
-    "jest": "*",
-    "jest-puppeteer": "*",
-    "puppeteer": "*",
-    "setup-polly-jest": "*"
+    "@pollyjs/adapter-puppeteer": "^4.0.2",
+    "@pollyjs/core": "^4.0.2",
+    "@pollyjs/persister-fs": "^4.0.2",
+    "jest": "^24.0.0",
+    "jest-puppeteer": "^4.0.0",
+    "puppeteer": "1.10.0",
+    "setup-polly-jest": "^0.6.0"
   }
 }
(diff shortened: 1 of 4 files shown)


examples/jest-puppeteer/package.json
```diff
@@ -5,15 +5,17 @@
...
   },
   "devDependencies": {
-    "@pollyjs/adapter-puppeteer": "*",
-    "@pollyjs/core": "*",
-    "@pollyjs/persister-fs": "*",
-    "jest": "*",
-    "jest-puppeteer": "*",
-    "puppeteer": "*",
-    "setup-polly-jest": "*"
+    "@pollyjs/adapter-puppeteer": "file:../../packages/@pollyjs/adapter-puppeteer",
```
offirgolan: Can the polly packages not just `*`? Would that prevent us from running build?

jasonmit: Not part of the workspace, would mean it would install them vs using the ones in the same project.  Adding the examples to the workspace though probably isn't worth it.  Cost of install goes up for something people probably rarely run (evident by how long they've been broken).

offirgolan: Yeah, but the initial point of these example is that they are contained and can be taken out of the polly repo and still work as expected to as closely imitate real life use-cases as possible.

jasonmit: I think we add a version range that we know is compatible with the example vs asterisk. This way we can update them at our leisure while keeping them stable.

offirgolan: Sounds good to me 😄

PR #306, opened 2020-02-01
ci: Adds back orphaned node adapter specifics test

### LINKED ISSUE #243: some integration tests are not being run
## Description

Per convo in #236, it looks like `test/integration/adapter-node-tests.js` is not being run and is most likely dead code. 

### Repro steps

Edit `test/integration/adapter-node-tests.js` and add e.g. `expect(false).to.equal(true);` into one of the `it` blocks.

jasonmit:
fixes #243


tests/integration/adapter-node-tests.js
```diff
@@ -4,35 +4,37 @@ export default function adapterNodeTests() {
   it('should handle recording requests posting a Buffer', async function() {
```
jasonmit: Looks like we forgot to bring this over when we removed adapter-fetch's node builds in favor of recommending adapter-node-http over it.

PR #310, opened 2020-02-20
Deprecates adapter & persister `name` in favor of `id`

### LINKED ISSUE #309: issue with `name` property when creating a custom adapter in TS
## Description

When creating a custom adapter using typescript, the name property conflicts with the `Function.name` constructor. not sure how to get around this, would it be possible to rename the property to `adapterName` or something similar? 

### Shareable Source

```js
class NewAdapter extends Adapter {
  // ...
  static get name() {
    return 'new-adapter';
  }
// ...
}
```

```
Static property 'name' conflicts with built-in property 'Function.name' of constructor function 'NewAdapter'.
```

### Dependencies

```json
{
  "@pollyjs/adapter": "^4.0.2"
}
```

## Relevant Links

https://github.com/microsoft/TypeScript/issues/442

## Environment

Node.js v12.6.0
darwin 18.7.0
yarn 1.19.1
npm 6.9.0

jasonmit:
Done in a backwards compatible way.  In the next major we can remove the deprecations.

Fixes #309

### DIFF
--- a/packages/@pollyjs/adapter-fetch/src/index.js
+++ b/packages/@pollyjs/adapter-fetch/src/index.js
@@ -8,7 +8,7 @@ const IS_STUBBED = Symbol();
 const REQUEST_ARGUMENTS = Symbol();
 
 export default class FetchAdapter extends Adapter {
-  static get name() {
+  static get id() {
     return 'fetch';
   }
 
--- a/packages/@pollyjs/adapter-fetch/src/index.js
+++ b/packages/@pollyjs/adapter-fetch/src/index.js
@@ -12,6 +12,12 @@ export default class FetchAdapter extends Adapter {
     return 'fetch';
   }
 
+  static get name() {
+    // NOTE: deprecated in 4.1.0 but proxying since it's possible "core" is behind
+    // and therefore still referencing `name`.  Remove in 5.0.0
+    return this.id;
+  }
+
   get defaultOptions() {
     return {
       context: global
--- a/packages/@pollyjs/core/src/-private/container.js
+++ b/packages/@pollyjs/core/src/-private/container.js
@@ -1,6 +1,6 @@
 import { assert } from '@pollyjs/utils';
 
-export function nameOfFactory(Factory) {
+export function idOfFactory(Factory) {
   if (Factory.hasOwnProperty('id')) {
     return Factory.id;
   }
@@ -13,10 +13,10 @@ export function nameOfFactory(Factory) {
 }
 
 function keyFor(Factory) {
-  return `${Factory.type}:${nameOfFactory(Factory)}`;
+  return `${Factory.type}:${idOfFactory(Factory)}`;
 }
 
-export default class Container {
+export class Container {
   constructor() {
     this._registry = new Map();
   }
@@ -33,7 +33,7 @@ export default class Container {
     );
 
     const { type } = Factory;
-    const name = nameOfFactory(Factory);
+    const name = idOfFactory(Factory);
 
     assert(
       `Invalid registration id provided. Expected string, received: "${typeof name}"`,
--- a/packages/@pollyjs/core/src/-private/container.js
+++ b/packages/@pollyjs/core/src/-private/container.js
@@ -1,7 +1,19 @@
 import { assert } from '@pollyjs/utils';
 
+export function nameOfFactory(Factory) {
+  if (Factory.hasOwnProperty('id')) {
+    return Factory.id;
+  }
+
+  console.warn(
+    `[Polly] ${Factory.id}:${Factory.name} "name" is deprecated and has been renamed to "id".`
+  );
+
+  return Factory.name;
+}
+
 function keyFor(Factory) {
-  return `${Factory.type}:${Factory.name}`;
+  return `${Factory.type}:${nameOfFactory(Factory)}`;
 }
 
 export default class Container {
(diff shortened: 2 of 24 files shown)


packages/@pollyjs/core/tests/unit/-private/container-test.js
```diff
@@ -27,16 +27,20 @@ describe('Unit | Container', function() {
...
     it('.register() - validation', function() {
-      class NoName extends Factory {
+      class NoId extends Factory {
         /* eslint-disable-next-line getter-return */
-        static get name() {}
+        static get id() {}
       }
 
       class NoType extends Factory {
+        static get id() {
```
jasonmit: masked exception since constructor.name was always a thing

tests/integration/adapter-tests.js
```diff
@@ -228,7 +228,8 @@ export default function adapterTests() {
     expect(resolved).to.have.members([1, 2, 3]);
   });
 
-  it('should work with CORS requests', async function() {
+  // NOTE: test very unstable because of typicode.com being down
+  it.skip('should work with CORS requests', async function() {
```
jasonmit: This test has been failing off and on for a couple months now, disabling.

packages/@pollyjs/persister-rest/src/index.js
```diff
@@ -4,10 +4,16 @@ import { buildUrl } from '@pollyjs/utils';
 import ajax from './ajax';
 
 export default class RestPersister extends Persister {
-  static get name() {
+  static get id() {
     return 'rest';
   }
 
+  static get name() {
```
jasonmit: All of these name getter proxies need to sit on the persister/adapter classes instead of base classes because:

```js
class Foo {
        static get name() {
            return 'foo';
        }
    }

    class Bar extends Foo {}

    Bar.name; // "Bar" not "foo"
```

tests/helpers/setup-fetch-record.js
```diff
@@ -29,9 +29,11 @@ setupFetchRecord.beforeEach = function(options = {}) {
 
 setupFetchRecord.afterEach = function() {
   afterEach(async function() {
-    this.polly.pause();
-    await this.fetchRecord({ method: 'DELETE' });
-    this.polly.play();
+    if (this.polly) {
```
jasonmit: Possible for test suite to fail before `this.polly` is set and made the stderr annoying to scroll through since we always assume it was.

packages/@pollyjs/core/src/polly.js
```diff
@@ -3,7 +3,7 @@ import { MODES, assert } from '@pollyjs/utils';
 import { version } from '../package.json';
 
 import Logger from './-private/logger';
-import Container from './-private/container';
+import Container, { nameOfFactory } from './-private/container';
```
offirgolan: iirc rollup has had issues supporting default and named exports. Might need to switch them both to a named export.

packages/@pollyjs/core/src/-private/container.js
```diff
@@ -1,7 +1,19 @@
 import { assert } from '@pollyjs/utils';
 
+export function nameOfFactory(Factory) {
```
offirgolan: Should this be `idOfFactory` or `factoryId`?

packages/@pollyjs/adapter-fetch/src/index.js
```diff
@@ -8,10 +8,16 @@ const IS_STUBBED = Symbol();
...
 export default class FetchAdapter extends Adapter {
-  static get name() {
+  static get id() {
     return 'fetch';
   }
 
+  static get name() {
+    // NOTE: deprecated in 4.1.0 but proxying since it's possible "core" is behind
+    // and therefore still referencing `name`.  Remove in 5.0.0
+    return this.id;
```
offirgolan: Should we just release a new major version instead of having to deal with all of these deprecations?

tests/helpers/setup-fetch-record.js
offirgolan: Might be good to add an inline comment about this.

packages/@pollyjs/core/src/polly.js
```diff
@@ -262,7 +262,7 @@ export default class Polly {
 
     if (typeof nameOrFactory === 'function') {
       container.register(nameOrFactory);
-      adapterName = nameOrFactory.name;
+      adapterName = nameOfFactory(nameOrFactory);
```
offirgolan: `adapterId`

packages/@pollyjs/adapter-fetch/src/index.js
jasonmit: It's a minor change imo and will help anyone with custom adapters to migrate without reading docs.

packages/@pollyjs/core/src/polly.js
jasonmit: and our tests still pass with this issue?

tests/helpers/setup-fetch-record.js
jasonmit: Done

packages/@pollyjs/core/src/-private/container.js
jasonmit: Done

packages/@pollyjs/core/src/polly.js
jasonmit: Done

Issue #269, opened 2019-10-24
Non-deterministic recordings files


brandon-leapyear:
## Description

I'm not sure how to minimally reproduce this, but using `persister-fs`, I'm getting non-deterministic orderings of requests. Context: I'm working on making our polly recordings idempotent; e.g. scrubbing timestamps and auth tokens. I have everything idempotent now, but some recordings (not all) keep changing the order of requests; e.g.
```
[
  <recording for /auth/login>,
  <recording for /users>,
  <recording for /users/4>,
]
```
will get reordered as
```
[
  <recording for /users/4>,
  <recording for /auth/login>,
  <recording for /users>,
]
```

It seems like `Persister` uses `stringify` from `fast-json-stable-stringify`, which should produce a deterministic JSON blob, but `Persister-FS` re-parses the JSON string as JSON, which makes the JSON blob non-deterministic again:

https://github.com/Netflix/pollyjs/blob/8f86fb372b6ac3e6efa36a9dee07d81d8e112847/packages/@pollyjs/persister-fs/src/index.js#L31

I also don't see a unit test for deterministic recordings?

https://github.com/Netflix/pollyjs/blob/master/packages/@pollyjs/persister-fs/tests/unit/persister-test.js

### Shareable Source

Haven't found a minimal repro yet

### Error Message & Stack Trace

See description

### Config

Copy the config used to setup the Polly instance:

```js
const { setupPolly } = require('setup-polly-jest')
setupPolly({
  adapters: ['node-http'],
  persister: 'fs',
  logging: false,
  recordFailedRequests: true,
  recordIfMissing: true,
  mode: recordMode,
  matchRequestsBy: {
    headers: {
      exclude: ['Authorization'],
    },
  },
})
```

### Dependencies

Copy the @pollyjs dependencies from `package.json`:

```json
{
    "@pollyjs/adapter-fetch": "^2.6.3",
    "@pollyjs/adapter-node-http": "^2.6.3",
    "@pollyjs/core": "^2.6.3",
    "@pollyjs/persister-fs": "^2.6.3",
    "@pollyjs/persister-rest": "^2.6.3",
}
```

## Environment

* Mac OS (darwin 18.7.0)
* Yarn: 1.17.3
* Node: v10.16.2

brandon-leapyear:
I wrote the following class as a workaround; it seems to do the job:
```js
import FSPersisterOld from '@pollyjs/persister-fs'
import fs from 'fs'
[...]

PR #321, opened 2020-03-31
feat(persister): Add `disableSortingHarEntries` option

offirgolan:
## Description

When disabled, entries in the the final HAR will be sorted by the request's timestamp.
This is done by default to satisfy the HAR 1.2 spec but can be enabled to improve
diff readability when committing recordings to git.

## Motivation and Context

Related to #269.

## Types of Changes

## Checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -241,6 +241,25 @@ polly.configure({
 });
 ```
 
+### disableSortingHarEntries
+
+_Type_: `Boolean`
+_Default_: `false`
+
+When disabled, entries in the the final HAR will be sorted by the request's timestamp.
+This is done by default to satisfy the HAR 1.2 spec but can be enabled to improve
+diff readability when committing recordings to git.
+
+**Example**
+
+```js
+polly.configure({
+  persisterOptions: {
+    disableSortingHarEntries: true
+  }
+});
+```
+
 ## timing
 
 _Type_: `Function`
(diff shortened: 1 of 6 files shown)


tests/integration/persister-tests.js
```diff
@@ -264,4 +264,85 @@ export default function persisterTests() {
...
+    await this.fetch(orderedRecordUrl(3));
+    await this.fetch(orderedRecordUrl(4));
+    await this.polly.persister.persist();
+
+    har = await this.polly.persister.find(recordingId);
+
+    expect(har).to.be.an('object');
+    expect(har.log.entries).to.have.lengthOf(4);
+    expect(har.log.entries[0].request.url).to.include(orderedRecordUrl(3));
+    expect(har.log.entries[1].request.url).to.include(orderedRecordUrl(4));
```
offirgolan: Due to the way we are adding entries, we add new ones to the beginning of the array and then keep only the unique ones (due to how `_.uniqWith` works), new entries will be at the top of the file instead of at the bottom.

docs/configuration.md
```diff
@@ -241,6 +241,25 @@ polly.configure({
 });
 ```
 
+### disableSortingHarEntries
```
offirgolan: Let me know if you have a better name for this option @jasonmit

jasonmit: I think this is a good name for the option 👍

tlands:
FWIW, the description of this option doesn't make sense...

> When disabled, entries in the the final HAR will be sorted by the request's timestamp

If I'm setting `disableSortingHarEntries: true` I would anticipate that the HAR file will _not_ be sorted. Can anyone clarify?

offirgolan:
@tlands you're right. I think the description is wrong. When disabled, it will **not** be sorted by timestamp.

PR #324, opened 2020-04-19
Make Adapter & Persister backwards compatible for implementors

geigerzaehler:
The replacement of the static `name` getter for `Adapter` and `Persister` in #310 broke custom implementations of these classes. To retain backwards compatibility we remove the requirement for `id` to be overridden.

## Motivation and Context

The changes in #310 required implementors of `Adapter` and `Class` to override the static `id` getter. This broke backwards compatibility with existing out-of-tree implementations of these classes—a simple patch upgrade would now throw exceptions.

## Description

To retain backwards compatibility we have `id` delegate to `name` by default. We restore the `name` getter and require an override of the method. This does not affect the in-tree implementations.

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -23,10 +23,15 @@ export default class Adapter {
     return 'adapter';
   }
 
-  /* eslint-disable-next-line getter-return */
   static get id() {
-    assert('Must override the static `id` getter.');
+    return this.name;
   }
+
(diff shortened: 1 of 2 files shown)


jasonmit:
Wouldn't this re-introduce https://github.com/Netflix/pollyjs/issues/309?

jasonmit:
https://github.com/Netflix/pollyjs/pull/325 should resolve this without re-introducing `name`

PR #325, opened 2020-04-19
fix: Legacy persisters and adapters should register

jasonmit:
Related to #324

### DIFF
--- a/packages/@pollyjs/adapter/src/index.js
+++ b/packages/@pollyjs/adapter/src/index.js
@@ -4,7 +4,8 @@ import {
   EXPIRY_STRATEGIES,
   PollyError,
   Serializers,
-  assert
+  assert,
+  getFactoryId
 } from '@pollyjs/utils';
 
 import isExpired from './utils/is-expired';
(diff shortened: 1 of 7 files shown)


packages/@pollyjs/utils/src/utils/id-of-factory.js
```diff
@@ -0,0 +1,13 @@
+export function idOfFactory(Factory) {
```
offirgolan, line 1: Thoughts of renaming this to `getFactoryId`?

packages/@pollyjs/utils/src/utils/id-of-factory.js
```diff
@@ -0,0 +1,13 @@
...
+  }
+
+  console.warn(
+    `[Polly] ${Factory.type}:${Factory.name} "name" is deprecated and has been renamed to "id".`
+  );
+
+  return Factory.name;
+}
+
+export default idOfFactory;
```
offirgolan, line 13: Why the need to re-export?

PR #329, opened 2020-04-29
fix(adapter-node-http): Improve binary response body handling

### LINKED ISSUE #327: The download file content is borken and the size looks doubled
## Description

I use Pollyjs with node-fetch and jest for my test.
The downloaded file content was not correct, the size looks doubled.
If I comment out the setupPolly part as the code bellow, the image downloaded is correct.
but if I un-comments the setuPolly, the image is not correct.

### Shareable Source

download.test.js
```js
const path = require('path');
const fetch = require('node-fetch')
const fs = require('fs')
const { Polly } = require('@pollyjs/core');
const { setupPolly } = require('setup-polly-jest');
const NodeHttpAdapter = require('@pollyjs/adapter-node-http');
const FSPersister = require('@pollyjs/persister-fs');

Polly.register(NodeHttpAdapter);
Polly.register(FSPersister);


describe('download image', () => {
/*
  setupPolly({
    adapters: ['node-http'],
    persister: 'fs',
    persisterOptions: {
      fs: {
        recordingsDir: path.resolve(__dirname, '../__recordings__')
      }
    }
  });
*/
  describe('from url', () => {
    it('should save to fs', async () => {
        let resp = await fetch("https://netflix.github.io/pollyjs/assets/images/wordmark-logo-alt.png")
        console.log(`content type: ${resp.headers.get('content-type')}`)
        console.log(`content length: ${resp.headers.get('content-length')}`)
        const dest = fs.createWriteStream('./placeholder.png')
        resp.body.pipe(dest)
    });
  }, 10000);

});
```

No

### Config

Copy the config used to setup the Polly instance:

```js
  setupPolly({
    adapters: ['node-http'],
    persister: 'fs',
    persisterOptions: {
      fs: {
        recordingsDir: path.resolve(__dirname, '../__recordings__')
      }
    }
  });
```

### Dependencies

```json
{
  "name": "pollytest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "jest"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@pollyjs/adapter-node-http": "^4.1.0",
    "@pollyjs/core": "^4.1.0",
    "@pollyjs/persister-fs": "^4.1.0",
    "jest": "^25.4.0",
    "setup-polly-jest": "^0.7.0"
  },
  "dependencies": {
    "node-fetch": "^2.6.0"
  }
}
```

## Relevant Links

No

## Environment

Node.js v13.12.0
darwin 19.4.0
npm: 6.14.4

offirgolan:
## Description

The core issue was in `getChunksFromBody` where we pass it the recorded body and response headers. Since the body is a string, it can't really know if the original buffer was binary or not. To get around that, we now store whether the response content is binary or not in the HAR file and expose it via the response class.

## Motivation and Context

Resolves #327.

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
+++ b/packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
@@ -125,26 +125,38 @@ function commonTests(transport) {
 
   it('should be able to download binary content', async function() {
     const url = `${protocol}//via.placeholder.com/150/92c952`;
-    const { server } = this.polly;
 
-    server.get(url).passthrough(true);
+    this.polly.disconnectFrom(NodeHTTPAdapter);
 
     const nativeResponseStream = await getResponseFromRequest(
       transport.request(url)
     );
 
-    server.get(url).passthrough(false);
+    this.polly.connectTo(NodeHTTPAdapter);
 
     const recordedResponseStream = await getResponseFromRequest(
       transport.request(url)
     );
 
-    const [nativeHash, recordedHash] = await Promise.all([
+    const { recordingName, config } = this.polly;
+
+    await this.polly.stop();
+    this.polly = new Polly(recordingName, config);
+    this.polly.replay();
+
+    const replayedResponseStream = await getResponseFromRequest(
+      transport.request(url)
+    );
+
+    const [nativeHash, recordedHash, replayedHash] = await Promise.all([
       calculateHashFromStream(nativeResponseStream),
-      calculateHashFromStream(recordedResponseStream)
+      calculateHashFromStream(recordedResponseStream),
+      calculateHashFromStream(replayedResponseStream)
     ]);
 
     expect(nativeHash).to.equal(recordedHash);
+    expect(recordedHash).to.equal(replayedHash);
+    expect(nativeHash).to.equal(replayedHash);
   });
 
   it('should handle aborting a request', async function() {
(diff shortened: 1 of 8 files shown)


packages/@pollyjs/adapter-node-http/tests/integration/adapter-test.js
```diff
@@ -125,26 +125,38 @@ function commonTests(transport) {
 
   it('should be able to download binary content', async function() {
     const url = `${protocol}//via.placeholder.com/150/92c952`;
-    const { server } = this.polly;
 
-    server.get(url).passthrough(true);
```
offirgolan, line 130: We weren't able to catch this issue since we were doing a passthrough which still went through the adapter.

packages/@pollyjs/persister/src/har/response.js
```diff
@@ -37,6 +37,10 @@ export default class Response {
 
     if (response.body && typeof response.body === 'string') {
       this.content.text = response.body;
+
+      if (response.isBinary) {
+        this.content._isBinary = true;
```
offirgolan, line 42: `_isBinary` is only going to the written to the har content if the content is binary. I didnt think it was necessary to always include it. @jasonmit thoughts?

jasonmit: seems fair but some what on topic - does this constitute a breaking change?

offirgolan: No, this is backwards compatible but does require to re-record to resolve the issue.

PR #342, opened 2020-06-04
fix(core): Compute order based on id and recording name

### LINKED ISSUE #340: Mocking server's downstream requests during parallel functional tests
## Description

Hello, I have a use case I'd like to share and hopefully get some insight into best practices. 

I'm running functional tests via Jest, i.e. spinning up my server, make requests to the server, assert the data. 

I'd like to record and replay mocks from the server process, so I'm spinning up the Polly instance on the server. I'm basically sending the test suite name with each request, and in this way I can assign a recordingName to Polly for each test:

```js
// src/index.ts

let polly;

app.use((req, res, next) => {
  const testName = req.headers['jest-functional-test-name'];
  const mode = req.headers['jest-functional-test-mocks-mode'] || 'replay';

  if (!polly && testName && mode !== 'passthrough') {
    polly = new Polly(testName, { ...pollyConfig, mode });
  }

  next();
});

app.post('/stop-polly', async (req, res) => {
  await polly.flush();
  await polly.stop();
  polly = null;
  return res.status(200).send();
});

```

This works if I run my test files in sequence. However, Jest, by default, runs test files in parallel, and it would be nice to leverage faster test runs (especially when running tests in record mode). 

Unfortunately, I'm having trouble recording and replaying mocks for tests that are running simultaneously and hitting the server at the same time. This is what I tried:

```js
// src/index.ts

const polly = new Polly('initialRecordingName', pollyConfig);

app.use((req, res, next) => {
  const testName = req.headers['jest-functional-test-name'];
  const mode = req.headers['jest-functional-test-mocks-mode'] || 'replay';

  if (mode === 'passthrough') {
    polly.passthrough();
  } else if (mode === 'record') {
    polly.record();
  } else if (mode === 'replay') {
    polly.replay();
  }

  if (testName) {
    polly.server
      .any()
      .filter((req) => {
        return req.headers['jest-functional-test-name'] === testName;
      })
      .recordingName(testName);
  }

  next();
});
```

Basically, I'm passing through the 'jest-functional-test-name' header to all downstream requests from my server, which allows me to use the `recordingName()` functionality along with filter to specify recordingName for each request based on the test from which it originates. 

This seems to work well for recording. However, replaying gets messed up, and I think it's bc of `order`. Let's say we run 2 tests, each of which request our server, which then for each requests makes to 2 requests to downstream services. The `order` value for these two requests should be 0 and 1 in each recording file, but unfortunately they end up being something like 0 and 2 in one file and 1 and 3 in the other file, bc I thnk order is being calculated "globally" as opposed to a single recordingName. Then when we replay, the test fails bc it can't find the correct record for the specific order. 

Does that make sense? It's a little hard for me to figure out exactly what's going on, but I've basically narrowed it down to an issue with order. 

I'm wondering if anyone has any thoughts/suggestions for this use case? I tried finding examples of this, but couldn't find anything.

### Dependencies

```json
{
  "@pollyjs/adapter-node-http": "^4.2.1",
  "@pollyjs/core": "^4.2.1",
  "@pollyjs/node-server": "^4.3.0",
  "@pollyjs/persister-fs": "^4.2.1",
}
```

## Environment

Node.js v10.19.0
darwin 18.7.0
npm 6.14.5

offirgolan:
## Description

Compute the request's order based on its id as well as the recording id it will be accessed from.

## Motivation and Context

Resolves #340.

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/core/src/-private/request.js
+++ b/packages/@pollyjs/core/src/-private/request.js
@@ -265,13 +265,14 @@ export default class PollyRequest extends HTTPBase {
     // Guid is a string representation of the identifiers
     this.id = md5(stringify(this.identifiers));
 
-    // Order is calculated on other requests with the same id
+    // Order is calculated on other requests with the same id and recording id
     // Only requests before this current one are taken into account.
     this.order =
       matchRequestsBy.order && !this.shouldPassthrough && !this.shouldIntercept
         ? requests
             .slice(0, requests.indexOf(this))
-            .filter(r => r.id === this.id).length
+            .filter(r => r.id === this.id && r.recordingId === this.recordingId)
+            .length
         : 0;
   }
 }
--- a/tests/integration/adapter-tests.js
+++ b/tests/integration/adapter-tests.js
@@ -3,28 +3,68 @@ import { ACTIONS } from '@pollyjs/utils';
 
 export default function adapterTests() {
   it('should respect request order', async function() {
-    let res = await this.fetchRecord();
+    const testOrder = async () => {
+      let res = await this.fetchRecord();
 
-    expect(res.status).to.equal(404);
+      expect(res.status).to.equal(404);
 
-    res = await this.fetchRecord({
-      method: 'POST',
-      body: JSON.stringify({ foo: 'bar' }),
-      headers: { 'Content-Type': 'application/json' }
+      res = await this.fetchRecord({
+        method: 'POST',
+        body: JSON.stringify({ foo: 'bar' }),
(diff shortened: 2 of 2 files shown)


offirgolan:
@jasonmit thoughts on this? This is potentially a breaking change since we are tweaking the way order is being calculated.

jasonmit:
> @jasonmit thoughts on this? This is potentially a breaking change since we are tweaking the way order is being calculated.

Agreed, changing existing functionality should be a breaking a change and I'm fine with it.

Issue #121, opened 2018-10-25
puppeteer adapter doesn't capture main frame


kiavashps:
I'm trying to use the puppeteer adapter to capture every request on a page (including the root html document, scripts, images, etc..) using:
```js
requestResourceTypes: [
    'document', 'stylesheet', 'image', 'media', 'font', 'script', 'texttrack',
    'xhr', 'fetch', 'eventsource', 'websocket', 'manifest', 'other'
]
```

When `requestResourceTypes` includes `document` (which matches the main frame/html page), the page hangs indefinitely (or hits timeout)

**Platform:** macOS v10.11

**Node.js:** v8.11.4

**@pollyjs/adapter-puppeteer:** v1.2.0
**@pollyjs/core:** v1.2.0
**@pollyjs/persister-fs:** v1.2.0
**puppeteer:** v1.9.0

```js
'use strict';

const {Polly} = require('@pollyjs/core');
const FSPersister = require('@pollyjs/persister-fs');
const PuppeteerAdapter = require('@pollyjs/adapter-puppeteer');
const puppeteer = require('puppeteer');

Polly.register(FSPersister);
Polly.register(PuppeteerAdapter);

(async () => {
    const browser = await puppeteer.launch({headless: false});
    const [page] = await browser.pages();

    await page.setRequestInterception(true);

    const polly = new Polly('hangs', {
        adapters: ['puppeteer'],
        adapterOptions: {
            puppeteer: {
                page,
                requestResourceTypes: [
                    'document'
                ]
            }
        },
        persister: 'fs',
        persisterOptions: {
            fs: {
                recordingsDir: `${__dirname}/recordings`
            }
        },
        recordFailedRequests: true,
        logging: true
    });

    await page.goto('http://www.example.com/', {
        waitUntil: 'networkidle2',
        timeout: 0
    });

    await page.close();
    await browser.close();
    await polly.stop();
})();
```

kiavashps:
Also, if I manually modify the `recording.har` file to add an entry for `https://www.example.com/`, it will correctly replay it. It's only when recording the request that it hangs.

offirgolan:
@kiavashps thanks for the detailed description on this issue. Does the page still hang if you set the resource types to `['xhr', 'fetch', 'document', 'other']`?

jasonmit:
Polly does not currently capture all of those resource types, only xhr, fetch, and other resource types.  Perhaps we should warn when other options are passed for `requestResourceTypes`.

[...]

Issue #345, opened 2020-06-19
[adapter-puppeteer] package is broken with Puppeteer 4 - removed prependListener and removeListener


Georgegriff:
## Description

Please read their release notes below before reading:

Puppeteer 4.0 removed its association with Node EventEmitter which this package relies upon https://github.com/puppeteer/puppeteer/releases/tag/v4.0.0

### Shareable Source
adapter-puppeteer relies on the following:prependListener which has been removed in Puppeteer.

It also relies on removeListener which has also been removed.
https://github.com/Netflix/pollyjs/blob/1223ba0dad16bf504074510ac060e7394a4cbac9/packages/%40pollyjs/adapter-puppeteer/src/index.js#L50
https://github.com/Netflix/pollyjs/blob/1223ba0dad16bf504074510ac060e7394a4cbac9/packages/%40pollyjs/adapter-puppeteer/src/index.js#L143

### Error Message & Stack Trace

```
COPY THE ERROR MESSAGE, INCLUDING STACK TRACE HERE
```

### Config

Copy the config used to setup the Polly instance:

```js
new Polly('Recording Name', {
  // config...
});
```

### Dependencies

Copy the @pollyjs dependencies from `package.json`:

```json
{
    "@pollyjs/adapter-puppeteer": "^4.0.4",
    "@pollyjs/core": "^4.0.4",
}
```

## Relevant Links

https://github.com/Netflix/pollyjs/blob/1223ba0dad16bf504074510ac060e7394a4cbac9/packages/%40pollyjs/adapter-puppeteer/src/index.js#L50
https://github.com/Netflix/pollyjs/blob/1223ba0dad16bf504074510ac060e7394a4cbac9/packages/%40pollyjs/adapter-puppeteer/src/index.js#L143

## Environment

Tell us which operating system you are using, as well as which versions of Node.js and npm/yarn. If applicable, include the browser and the corresponding version.

Run the following to get it quickly:

```
node -e "var os=require('os');console.log('Node.js ' + process.version + '\n' + os.platform() + ' ' + os.release())"
npm --version
yarn --version
```

arnaugm:
Hi! I'm interested in providing a fix for this issue, any guidance?

PR #348, opened 2020-07-02
feat: Simplify PuppeteerAdapter to avoid creating duplicate requests

lguychard:
## Description

Simplifies the PuppeteerAdapter to avoid creating "duplicate" requests for passthrough requests with a `pollyjs_passthrough_req_id` query parameter. This is changed to simply call `.continue()` on passthrough requests.

This also fixes the hanging issue when including `document` in `requestResourceTypes` (https://github.com/Netflix/pollyjs/issues/121#issue-374093084)

## Motivation and Context

I wanted to be able to record the initial `document` using the PuppeteerAdapter. I noticed that the reason this was not possible was that the `fetch()` request constructed by the adapter hung after receiving an `OPTIONS` response when requesting the initial document. I investigated removing these `fetch()` calls, and simply calling `request.continue()`, which would remove the hanging problem, and simplify the internal implementation.

## Types of Changes

## Checklist

- [x] I have added tests to cover my changes ([relevant tests pass](https://travis-ci.org/github/Netflix/pollyjs/jobs/704355655#L1280-L1296), the functionality of the Puppeteer adapter is otherwise preserved).

### DIFF
--- a/packages/@pollyjs/adapter-puppeteer/src/index.js
+++ b/packages/@pollyjs/adapter-puppeteer/src/index.js
@@ -1,9 +1,7 @@
 import Adapter from '@pollyjs/adapter';
-import { URL } from '@pollyjs/utils';
 
 const LISTENERS = Symbol();
 const PASSTHROUGH_PROMISES = Symbol();
-const PASSTHROUGH_REQ_ID_QP = 'pollyjs_passthrough_req_id';
 
 export default class PuppeteerAdapter extends Adapter {
   static get id() {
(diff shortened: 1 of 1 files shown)


offirgolan:
@lguychard thank you for taking the time to work on this PR.

Unfortunately, the 2nd request with the `PASSTHROUGH_REQ_ID_QP` query parameter is necessary since we don't want to respond to the original request until the polly instance is done interacting with it. Since the request and response can be modified at anytime during the lifecycle of the request via the supported events or full interception, we can't just `continue` the original request.

To be honest, the entire adapter needs to be rethought out because of #345 as well as some other issues we've seen with xhr vs fetch requests.

lguychard:
Thanks for taking a look @offirgolan 

> full interception

 As far as full interception goes, this PR leaves the existing functionality unchanged AFAICT.

> the request and response can be modified at anytime during the lifecycle of the request

This is indeed a use case I had overlooked, thanks for pointing it out! Correct me if I'm wrong, but it seems like this would specifically apply to `beforeResponse` and `response` events.

A possible path forward there would be to use the devtools protocol's [`Fetch` API](https://chromedevtools.github.io/devtools-protocol/tot/Fetch). There is previous art at https://github.com/jsoverson/puppeteer-interceptor, and this would still allow to `continue()` the original request and to avoid issuing a separate `fetch()` request.

offirgolan:
> Correct me if I'm wrong, but it seems like this would specifically apply to beforeResponse and response events.

Correct. The `request` event can also modify the final outgoing request to be made (e.g. changing query params or post body).

> A possible path forward there would be to use the devtools protocol's Fetch API. 

That would be great. I wonder if we can use jsoverson/puppeteer-interceptor directly to handle all the interception logic.

lguychard:
Closing this as it won't be merged as-is. We (Sourcegraph) are working internally on a new adapter using CDP Fetch, and will upstream it when it's ready (cc @marekweb).

offirgolan:
That's really awesome to hear! Please don't hesitate to reach out if you need some eyes on it.

felixfbecker:
Thanks for the offer @offirgolan! The PR is here: https://github.com/sourcegraph/sourcegraph/pull/13870, would love to get your review!

PR #420, opened 2021-11-28
feat!: Use base64 instead of hex encoding for binary data

### LINKED ISSUE #371: Polly should use standard HAR `encoding` property with Base64, not non-standard `_isBinary` with hex content (4x the size)
## Description

Polly's support for binary contents (e.g. images) currently works by writing `_isBinary: true` into the HAR and encoding the body in hexadecimal. Polly uses this as an internal representation too, which is a bit weird since internally the content can just be a much more efficient `Buffer`.

This is non-standard. It means that a HAR from Polly is not interoperatable with other tools like Chrome dev tools, e.g. it is impossible to load a Polly HAR in Chrome because Chrome will treat the hex content as a literal string.

Hexadecimal encoding is also extremely large - 4x (300%!) as Base64, and Base64 is already 37% larger than binary. Polly HARs are committed to Git, so this is a problem even much larger than committing raw binary into Git repositories (increased clone time that can never be reduced again).

There is a standard in HAR to encode binary files, [the `encoding` field](https://w3c.github.io/web-performance/specs/HAR/Overview.html#content:~:text=encoding%20%5Bstring%2C%20optional%5D%20(new%20in%201.2),its%20original%20character%20set%20into%20UTF%2D8):

> encoding [string, optional] (new in 1.2) - Encoding used for response text field e.g "base64". Leave out this field if the text field is HTTP decoded (decompressed & unchunked), than trans-coded from its original character set into UTF-8.

This can be set to `base64` to treat the content as Base64, which would be interoperatable with other tools and 1/4th of the size of hexadecimal encoding.

offirgolan:
Resolves #371

### DIFF
--- a/packages/@pollyjs/adapter-fetch/src/index.js
+++ b/packages/@pollyjs/adapter-fetch/src/index.js
@@ -190,8 +190,8 @@ export default class FetchAdapter extends Adapter {
     return {
       statusCode: response.status,
       headers: serializeHeaders(response.headers),
-      body: buffer.toString(isBinaryBuffer ? 'hex' : 'utf8'),
-      isBinary: isBinaryBuffer
+      body: buffer.toString(isBinaryBuffer ? 'base64' : 'utf8'),
+      encoding: isBinaryBuffer ? 'base64' : undefined
     };
   }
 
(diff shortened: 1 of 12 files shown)


roschaefer:
@offirgolan @netflixgithub could you please release `@pollyjs/adapter-node-http` and the other packages affected by this PR? The last time it was `Published a year ago`.

I'm trying to decode a `_isBinary` response object in an `beforePersist` hook. Kind of similar to what @renanvalentin asked for [here](https://github.com/Netflix/pollyjs/issues/371#issuecomment-751740811). Is there a way to always save a response as `utf-8` (avoid `_isBinary`) in order to filter sensitive data before persisting?

roschaefer:
Yep, my bad. I was looking at the `@types/pollyjs__adapter-node-http` package. Thanks :+1:

PR #441, opened 2021-12-18
fix(persister): `requests` -> `request` in `HarEntry` declaration

ROpdebee:
## Description

Fix typo in `HarEntry` type declaration. In reality, [the property is called `request`](https://github.com/Netflix/pollyjs/blob/18262cf3213d3f3ca7a778a574c162dc2318860a/packages/%40pollyjs/persister/src/har/entry.js#L18) but the declaration uses `requests` instead.

## Motivation and Context

See above.

In addition to this fix, the declaration is also lacking the `_recordingName` and `creator` properties for `HarLog` instances, but I'm not sure how to best handle these as they are added by the base persister implementation, so it could be possible that a custom persister omits these properties. Perhaps the `HarLog` declaration should be made generic to handle this? Happy to add it to this PR in case that's desired.

## Types of Changes

## Checklist

### DIFF
--- a/packages/@pollyjs/persister/types.d.ts
+++ b/packages/@pollyjs/persister/types.d.ts
@@ -39,7 +39,7 @@ export interface HarEntry {
   _id: string;
   _order: number;
   startedDateTime: string;
-  requests: HarRequest;
+  request: HarRequest;
   response: HarResponse;
   cache: {};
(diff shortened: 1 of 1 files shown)


offirgolan:
Thanks for the PR @ROpdebee!

